PR #1, opened 2019-12-17
Feat/basic simulations

giovannic:
A tested for creating simulations. At the moment it can handle states and stateless processing functions.

The next steps are to:
 1. add variables
 2. copy over the malaria model from my prototype
 3. test the malaria processing functions


R/individual.R
```diff
@@ -0,0 +1,57 @@
+# Class: Individual
+# Represents an individual in our simulation
+# `name` is a unique idetifier which is used in the output
+# `states` a list of State objects
+setClass(
+  'Individual',
+  slots = list(name='character', states='list'),
+  validity = function(object) {
+    names <- sapply(object@states, function(state) { state@name })
```
richfitz: Avoid `sapply` as it's not type-safe

```r
sapply(list(1:5, 1:10), mean) # 3, 5.5
sapply(list(1:5, 1:10)[0], mean) # list()
```

use `vapply` and provide a return type:

```r
vapply(object@states, function(state) state@name, character(1))
```

R/individual.R
```diff
@@ -0,0 +1,57 @@
...
+    if (any(duplicated(names))) {
+      return('No duplicate states allowed')
+    }
+    T
```
richfitz: Avoid T and F as they're not reserved symbols and just bound to `TRUE` and `FALSE`

docker/development/Dockerfile
```diff
@@ -6,3 +6,7 @@ RUN apt-get update && apt-get install -y \
   libxml2-dev
 
 RUN R -e "install.packages(c('devtools', 'roxygen2', 'testthat', 'knitr'))"
```
richfitz: r-base has been a pain to depend on as a source image and we use `rocker/r-ver` (though that is also a pain because you have disable the pinned MRAN repo - see [here](https://github.com/vimc/orderly/blob/617e16ab8cf1c798d9527b4fcc87ed6fbe789745/docker/Dockerfile#L10)). The r-base image is built off a peculiar set of debian images and you will periodically get errors saying that you are "holding broken packages", and will be unable to update your images until something changes upstream - this will disrupt you 2-3 times a year.

tests/testthat.R
```diff
@@ -1,4 +1,6 @@
 library(testthat)
+library(testutils)
```
richfitz: where is this and what does it do?

DESCRIPTION
```diff
@@ -1,13 +1,18 @@
...
+           role = c('aut', 'cre'),
+           email = 'giovanni.charles@gmail.com'
            )
-Description: What the package does (one paragraph).
+Description: Individual gives you useful primitives for structuring and simulating individual based epidemiological models.
 License: None
 Encoding: UTF-8
 LazyData: true
+Imports:
+    dplyr,
```
richfitz: These packages are only used in tests and should go into `Suggests:`, not into `Imports:`

tests/testthat/test-simulation-e2e.R
```diff
@@ -1,46 +1,84 @@
...
+  R <- State('R', 0)
+  human <- individual::Individual('human', S, I, R)
+  simulation <- individual::simulate(human, list(), 4)
+  true_df <- data.frame(
+    timestep=as.numeric(rep(0:4, 1, NA, 4)),
+    state=factor(c(
+      rep('S', 20)
+    ), levels = c('S', 'I', 'R'))
+  )
+  message <- dplyr::all_equal(true_df, simulation$render(human))
```
richfitz: Can you please not depend on dplyr for this? If you just want to test almost-equal, then `expect_equal` will do this for you

tests/testthat/test-simulation-e2e.R
```diff
@@ -1,46 +1,84 @@
...
+  expect(
```
richfitz: This is an odd construction - can you not do:

```
expect_mapequal(simulation$render(human), true_df)
```

and elsewhere within the test suite. Directly using `expect` should be quite rare

tests/testthat/test-simulation-e2e.R
```diff
@@ -1,46 +1,84 @@
...
-            to,
-            i[1:min(rate,length(i))]
-          )
-        )
+    return(function(frame) {
+      from_state <- get_state(frame, human, from)
+      individual::StateUpdate(
+        human,
+        to,
+        from_state[1:min(rate,length(from_state))]
```
richfitz: avoid sequences of the form `1:n` because you get the unexpected result with `n` as `0` - use `seq_len` instead

R/simulate.R
```diff
@@ -0,0 +1,134 @@
...
+      )
+    }
+  )
+  new('SimFrame', frames = setNames(frames, names))
+}
+
+# Class: Simulation
+# Representation class for a list of simframes spanning a timescale
+# `sim_frames` is a list of (SimFrame, timestep) pairs
+Simulation <- setRefClass(
```
richfitz: Related to the S4 objection above - if the reason for using S4 is that you want reference classes to get Python/Java style classes, please use [R6](https://cran.r-project.org/web/packages/R6/index.html)

docker/development/Dockerfile
```diff
@@ -6,3 +6,7 @@ RUN apt-get update && apt-get install -y \
...
+
+COPY . /home/docker/individual
+
+RUN R -e "library('devtools'); install_deps('/home/docker/individual')"
```
richfitz: you should be able to get this a lot lighter by using `remotes` rather than in dragging devtools

DESCRIPTION
```diff
@@ -1,13 +1,18 @@
...
 Authors@R:
     person(given = "Giovanni",
-           family = "Charles"
+           family = "Charles",
+           role = c('aut', 'cre'),
+           email = 'giovanni.charles@gmail.com'
            )
-Description: What the package does (one paragraph).
+Description: Individual gives you useful primitives for structuring and simulating individual based epidemiological models.
 License: None
```
richfitz: Can you add a licence please? Note that MIT/BSD etc have a special and peculiar format - if you do `usethis::use_mit_licence()` it will do the dirty work for you. Our previous reading of the contract is that we must assign Imperial as the copyright holder (goes in Authors@R too) `Imperial College of Science, Technology and Medicine`

R/individual.R
```diff
@@ -0,0 +1,57 @@
+# Class: Individual
+# Represents an individual in our simulation
+# `name` is a unique idetifier which is used in the output
+# `states` a list of State objects
+setClass(
```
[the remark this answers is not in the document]
giovannic: _looks up R6 classes_ 😮 

right away!

The "Advanced R" book I was using is 2015

docker/development/Dockerfile
giovannic: do you mean this? https://github.com/r-lib/remotes

Am I right in thinking I'd have to push my DESCRIPTION file to github before I could install it locally then? Because that sounds inconvenient.

I do agree that devtools is an annoying installation when it's not cached in my docker. But we need it to run tests and package building stuff. How would I get rid of it?

richfitz: Yup, that's the package. `devtools` actually uses it for `install_deps` so you can do `remotes::install_deps('/home/docker/individual')` and replace`RUN R -e "install.packages(c('devtools', 'roxygen2', 'testthat', 'knitr'))"` with just installing `remotes`.

What we typically do is to duplicate the dependency list directly in the Dockerfile (e.g., https://github.com/vimc/orderly/blob/617e16a/docker/Dockerfile#L13-L28) which is not beautiful but results in a somewhat smaller image, fewer layers, and faster build times (because `COPY` will invalidate the build cache, it's nice to do that after the dependencies are installed)

tests/testthat/test-simulation-e2e.R
giovannic: Is there an `expect` for dataframes that allows an independent ordering of rows and columns?

giovannic: it's an odd construction because all_equal returns TRUE if the frames are equal but a truthy string if they are not 🤦‍♂

DESCRIPTION
giovannic: http://r-pkgs.had.co.nz/description.html#license has a slightly nicer format, would that do?

docker/development/Dockerfile
giovannic: why do we have to disable mran? Microsoft's spiel seems convincing and it seems to be working

DESCRIPTION
richfitz: Sorry, I don't follow

docker/development/Dockerfile
richfitz: Keep it in for now if you're convinced - the issue is that it pins to a particular day of MRAN (so for r-ver:3.6.1 it's 2019-12-06, which was I think the release date of 3.6.2) - that means you silently don't get updated versions of packages.  I have found this to be more disruptive than useful but YMMV

DESCRIPTION
giovannic: I misread your comment. Done, thanks!

.gitignore
```diff
@@ -1 +1,7 @@
 *.sw*
+.RData
+.rhistory
+individual.Rcheck
+individual_*.tar.gz
+.Rd2pdf*
+man
```
richfitz: Even though these are generated files, we generally commit them to the repository as it means that the package can be directly installed without needing roxygen2. You're welcome to go your own way on this of course, but be aware that it is unusual

DESCRIPTION
```diff
@@ -1,13 +1,31 @@
...
+    given = "Imperial College of Science, Technology and Medicine",
+    family = "",
+    role = c('cph'),
+    email = 'giovanni.charles10@imperial.ac.uk'
+  ))
+Description: This package gives you useful primitives for structuring and simulating individual based epidemiological models.
+License: MIT + file LICENSE
 Encoding: UTF-8
 LazyData: true
+Collate:
```
richfitz: The collate section can be removed now I believe - possibly vestigial from the S4 implementation?

R/utils.R
```diff
@@ -0,0 +1,74 @@
+#' Creates a DataClass design pattern
+#' readonly class for collecting related fields
+#' implementation inspired by https://github.com/r-lib/R6/issues/186
+#' @param classname is the name of the class
+#' @param fields is a sequence of readonly fields to create
+#' @param initialize is the initialisation funtion
+#' @param print_fields are the fields to print
+DataClass <- function(classname, fields, initialize, print_fields = NULL) {
```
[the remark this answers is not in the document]
giovannic: This is really cool.

Although it does sacrifice encapsulated R6 methods, like Individual's `check_state`. It also introduces a new type of object to the package, so we'd have environments as well as R6 objects, and no consistent way to initialise them.

Is there something wrong with active bindings?

DESCRIPTION
giovannic: This is actually an unfortunate sideeffect of creating high-level functions like DataClass or readonly_accessor. Apparently they can't be lazily loaded when defining `State`, `Individual`, ...

R/utils.R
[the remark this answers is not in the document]
giovannic: Completely separately I think your general comment about a using growing vector was quite important. I've made a change to using a 3D array that is allocated at the beginning of the simulation. I think it was worth getting that right from the beginning.

In the latest update I also added a simple performance benchmark to check the improvement. There were large improvements in space and time.

PR #2, opened 2020-01-06
Feat/variables

giovannic:
Adding variables and constants to the simulation


R/individual.R
```diff
@@ -51,3 +79,45 @@ State <- DataClass(
...
+    private$.name <- name
+    private$.initialiser <- initialiser
+  },
+  print_fields = c('name')
+)
+
+#' Class: Constant
+#' Represents a constant for an individual in our simulation
+#' @export Constant
+Constant <- DataClass(
```
richfitz: I'm not following the difference here between `Variable` and `Constant` objects - they seem identical (down to the typo in `idetifier`) so the only difference is what they're inserted into within an `Individual`?  (along those lines, the unique restriction in the `@param name` seems more a property within `Individual` than of a `Variable`?). Is more planned that distinguishes these (update methods etc?) or should they be merged?

R/simulation.R
```diff
@@ -0,0 +1,176 @@
...
+    #' Get a SimFrame for the current timestep
+    get_current_frame = function() {
+      SimFrame$new(
+        private$.individuals,
+        lapply(private$.individuals, function(i) {
+          private$.individual_to_states[[i$name]][,private$.current_timestep]
+        }),
+        lapply(private$.individuals, function(i) {
+          private$.individual_to_variables[[i$name]][
+            ,,private$.current_timestep, drop=FALSE
```
richfitz: I hope you found out about `drop = FALSE` without too much frustrating debugging

R/simulation.R
```diff
@@ -0,0 +1,176 @@
...
+        private$.individual_to_variables[[name]][
+          ,,private$.current_timestep + 1
+        ] <- private$.individual_to_variables[[name]][
+          ,,private$.current_timestep
+        ]
+      }
+
+      # execute process updates
+      for (update in updates) {
+        if (class(update)[1] == 'StateUpdate') {
```
richfitz: Will this always be the first class value? Or should this be `inherits(class, "StateUpdate")`?

R/frame.R
```diff
@@ -0,0 +1,86 @@
...
+    #' @description
+    #' Get the index of individuals with a particular state
+    #' @param individual of interest
+    #' @param state of interest
+    get_state = function(individual, ...) {
+      states <- list(...)
+      if (!(individual$name %in% names(private$.states))) {
+        stop('Unregistered individual')
+      }
+      valid_states <- all(
```
richfitz: Any reason not to do:

```
for (state in states) {
  if (!individual$check_state(state)) {
    stop('Invalid state')
  }
}
```

(replacing 20-29)

R/frame.R
```diff
@@ -0,0 +1,86 @@
...
+          states,
+          function(state) individual$check_state(state),
+          logical(1)
+        )
+      )
+      if (!valid_states) {
+        stop('Invalid state')
+      }
+      individual_frame <- private$.states[[individual$name]]
+      which(
```
richfitz: I think that this `which(... %in% ...)` construct can be replaced by `match`, which is what `%in%` is implemented on top of

R/simulation.R
```diff
@@ -0,0 +1,176 @@
...
+      individual_names <- lapply(individuals, function(i) { i$name })
+      states <- lapply(individuals, function(i) {
+        population <- sum(
+          vapply(i$states, function(s) s$initial_size, numeric(1))
+        )
+        a <- array(
+          rep(NA, population * timesteps),
+          c(population, timesteps)
+        )
+        a[,1] <- unlist(
```
richfitz: Perhaps

```
a[,1] <- rep(
  vapply(i$states, function(x) x$name, character(1)),
  vapply(i$states, function(x) x$initial_size, numeric(1)))
```

(`unlist(lapply(...))` always makes me a little uneasy)

R/simulation.R
```diff
@@ -0,0 +1,176 @@
...
+  if (end_timestep <= 0) {
+    stop('End timestep must be > 0')
+  }
+  if (! is.list(individuals)) {
+    individuals <- list(individuals)
+  }
+  output <- Simulation$new(individuals, end_timestep)
+  frame <- output$get_current_frame()
+  for (timestep in seq_len(end_timestep - 1)) {
+    updates <- unlist(
```
richfitz: I don't think this `unlist` is doing anything? At least not for the 2nd test in simulation-e2e

R/simulation.R
giovannic: It's all a distant memory haha

R/frame.R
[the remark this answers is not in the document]
giovannic: I _am_ worried about the performance impact of these lookups. Yes, I would like that issue to disappear in the C++ translation.

For the time being, it's quite useful. The checks have stopped me making a couple of mistakes on the malaria model!

giovannic: unfortunately match only returns the first occurance :(

R/simulation.R
giovannic: I've made an issue to write some extra e2e tests to make sure this is doing what I'm expecting

giovannic:
Thanks @richfitz, super helpful stuff!

PR #4, opened 2020-02-11
Add documentation:

giovannic:
* Add readme
* Add example for simulate
* Add vignette for modelling
* Unignore man pages and static site


giovannic:
Based on the last code review, it seemed like this package was becoming a bit abstract.

Here I add a readme, some more documentation and a vignette on modelling.

Keen to read your thoughts!

vignettes/Modelling.Rmd
```diff
@@ -0,0 +1,218 @@
...
+
+```{r}
+simulation <- simulate(human, processes, 100)
+rendered <- simulation$render(human)
+
+plot_states <- function(output) {
+  df <- rbind(
+    data.frame(
+      counts = colSums(rendered$states == 'S'),
+      timestep = 1:100,
```
richfitz:
```
seq_len(ncol(rendered$states))
```

rather than the magic number here? Or would it be worth (at some point) having time retuerned from the `render()` method?  That might be particularly useful if the state gets thinned too

vignettes/Modelling.Rmd
```diff
@@ -0,0 +1,218 @@
...
+affected by the environment and can in turn affect state changes.
+
+Let’s take immunity for example. We could decide that it should be acquired when
+individuals individuals recover from infection and should reduce the chances of
+immune individuals from becoming re-infected.
+
+We would first want to update our model structure…
+
+```{r}
+immunity <- Variable$new('immunity', function(size) 0)
```
richfitz: What is the meaning of `size` here? And is 0 the initial value?

vignettes/Modelling.Rmd
```diff
@@ -0,0 +1,218 @@
...
+```
+
+Here we initialise all levels of immunity to 0.
+
+We can then start to describe how this affects our processes...
+
+```{r}
+infection <- function(frame, timestep, parameters) {
+  from_state <- frame$get_state(human, S)
+  rate_modifier <- 1 - frame$get_variable(human, immunity)[from_state]
```
richfitz: Is immunity here applying to the entire population?

richfitz:
Timely help from Gabor: https://twitter.com/GaborCsardi/status/1227568604289302528

vignettes/Modelling.Rmd
giovannic: I have added a comment to make this line a bit more readable. But it could be a good idea to add a `subset` parameter to the API 🤔

vignettes/Modelling.Rmd
giovannic: The example now makes it clear why size is needed

giovannic:
> 
> 
> Timely help from Gabor: https://twitter.com/GaborCsardi/status/1227568604289302528

This was very helpful

PR #6, opened 2020-03-16
Feat/speed

giovannic:
So this is the translation of the simulation code to c++. It also includes some changes to clean up obsolete design decisions. As a summary:

 * Simulation and SimFrame are in cpp
 * `DataClass`es have been removed
 * A `Render` class has been added to save clients from storing the whole simulation in memory.

Notes for contention:

 1. Rcpp modules. It would be good to get your opinion on this. It still seems to be the way to expose classes and it makes the interface code very concise. This comes at the cost of a bigger compilation time and size (I think libs is now 4.8Mb). Despite this, I found it quite manageable while developing so I didn't feel any urgency to translate to Rcpp Xptrs.

2. R6 in cpp. The original API made clients specify models and processes using R6 objects. It read a nicer than passing vague, unvalidated named lists around. In translating the minimum to cpp, these classes: `Individual`, `State`, `Variable`, `*Update` are still in R. Instead there is a little code for dealing with reading R6 objects. There would be little to no performance improvement from translating these classes to cpp but perhaps a small readability improvement.

The reviewer's thoughts on these two points would be useful.


src/types.h
```diff
@@ -0,0 +1,28 @@
...
+#ifndef SRC_TYPES_H_
+#define SRC_TYPES_H_
+
+#include <memory>
+#include <valarray>
+#include <vector>
+#include <unordered_map>
+#include <unordered_set>
+
+using namespace std;
```
richfitz: IME this always seems to cause problems in the long run for the cost of writing `std::` all over the show. Would you mind not doing it in the headers at least?  Same with `Rcpp::` and the usage in other headers

src/Simulation.cpp
```diff
@@ -0,0 +1,179 @@
...
+            apply_variable_update(update, new_variables);
+        } else {
+            stop("Unknown update type");
+        }
+    }
+}
+
+void Simulation::apply_state_update(const Environment update, states_t& new_states) {
+    Log(log_level::debug).get() << "updating state" << endl;
+    const auto individual_name = nested_accessor<string>(update, {"individual", "name"});
```
richfitz: Is this sort of thing going to be fast enough in general? There's quite a bit of basic R types here (List, Environment and some vector types) and I wonder if some of these bits of accessing can be done once and not within things that are going to end up in tight loops.  Something to quickly repack things into a std::map or equivalent so that we get all the R searching over and done with.  I guess it's an unavoidable cost though, but if we can avoid paying it every timestep that might be worth while

Something to ponder on I guess.  Can you point me at where this is used within the malaria model perhaps?

Two things that might be useful (I've used neither)

* https://r-prof.github.io/jointprof/articles/jointprof.html
* https://twitter.com/unconj1/status/1240341055503060995

giovannic: 🤔 it's a good point.

Here's a nice isolated [example](https://github.com/mrc-ide/malariasimulation/blob/dev/R/mosquito_emergence.R) of a state update in the malaria code.

I hadn't actually managed to profile the C++ code. Jointprof looks marvellous!!

If it turns out that `nested_accessor` is expensive, I could look into having a separation between `Individual`s and `IndividualHandle`s. The `Handle` (probably just a string) would be passed into processes so that they don't have to be unpacked.

Will bear this in mind for the coming rounds of optimisation. However, the bottleneck at the moment is in `get_state` and the `render$update` (which is still in R). Should open up `apply_updates` soon after!

Screenshot 2020-03-26 at 12 33 01

giovannic:
So I've qualified std, Rcpp and added .dll. I've responded to the `nested_accessor` convo.

Issue #9, opened 2020-05-29
Optimise data structure for `individual_index_t`


giovannic:
Insertions and iterations to hasmaps are two of the main bottlenecks at the moment. See the latest [profile](https://trello.com/c/AwBABG8t/88-speed-up-round-2)

PR #13, opened 2020-06-09
Feat/ctzl linear

giovannic:
A new data structure for storing the state of an individual:

Pros:
 * Big speed improvements on insertion and erasure (image included)
 * Big speed improvements on malaria model (image included)
 * Opens up the possibility for parallelism (unordered_set is not thread safe)

Caveats:
 * Slower iteration over individuals in a state
 * Limits compatability to recent processors and the GCC compiler because of [ctzl instruction](https://github.com/mrc-ide/individual/compare/feat/ctzl_linear?expand=1#diff-723a88bba0330a7750750dc131a95101R90)
 * Probably most performant on 64bit processors
 * Marginally larger memory usage.

Closes [this](https://github.com/mrc-ide/individual/issues/9)

-----profiling results-----
# Unordered set
unordered_set

# Bitset
bitset

# Microbenchmark comparison
Screenshot 2020-06-09 at 16 47 09


inst/include/IndividualIndex.h
```diff
@@ -0,0 +1,223 @@
...
+ *  Created on: 1 Jun 2020
+ *      Author: gc1610
+ */
+
+#ifndef INST_INCLUDE_INDIVIDUALINDEX_H_
+#define INST_INCLUDE_INDIVIDUALINDEX_H_
+
+#include <cmath>
+
+class IndividualIndex;
```
richfitz, line 13: Given the impact that this has and the density of the code below, it would be worth including a short comment explaining what this is for, equivalent to, and a reference to the underlying theory, even if just a wikipedia page etc

inst/include/IndividualIndex.h
```diff
@@ -0,0 +1,223 @@
...
+    auto excess = p % num_bits;
+    uint64_t bitset = bitmap.at(bucket) >> excess;
+
+    while(bitset == 0 && bucket + 1 < bitmap.size()) {
+        bitset = bitmap.at(++bucket);
+        excess = 0;
+    }
+
+    auto lsb = bitset & -bitset;
+    auto r = __builtin_ctzll(lsb);
```
richfitz, line 90: As discussed we should make this swappable for a non-builtin version where not available - that should not be a proble,.

Is there any chance that `lsb` will be zero here? If so the result is undefined

inst/include/Scheduler.h
```diff
@@ -10,33 +10,37 @@
 
 #include "common_types.h"
 
-using event_t = std::pair<std::string, std::vector<SEXP>>;
+using event_t = std::tuple<std::string, size_t, std::vector<SEXP>>;
```
richfitz, line 13: Could you document what the 3 tuple elements are please?

inst/include/Scheduler.h
```diff
@@ -56,31 +60,35 @@ inline void Scheduler<TProcessAPI>::process_events(
...
+                    (*cpp_listener)(*cpp_api, target);
                 } else {
                     Rcpp::Function r_listener = listener;
-                    const auto& target = timeline.at(current_timestep);
-                    r_listener(
-                        r_api,
-                        std::vector<size_t>(std::cbegin(target), std::cend(target))
-                    );
+                    auto r_target = std::vector<size_t>(target.size());
+                    auto i = 0;
```
richfitz, line 75: Sorry, tiny thing, but why does this need to be done outside of the loop if it's just used within the loop?

inst/include/IndividualIndex.h
```diff
@@ -0,0 +1,223 @@
...
+ */
+
+#ifndef INST_INCLUDE_INDIVIDUALINDEX_H_
+#define INST_INCLUDE_INDIVIDUALINDEX_H_
+
+#include <cmath>
+
+class IndividualIndex;
+
+class IndividualIndex {
```
richfitz, line 15: Is it worth templating this over a type? Would we ever want something smaller than size_t here? Does that _work_ with the lower-level operations you have here? Would we get any better performance with `IndividualIndex<int32_t>` if it was implemented and if we only needed 4 billion individuals?

inst/include/Scheduler.h
giovannic: Here, if `i` was declared inside the loop it would be the a new `i` each time, so I couldn't increment it.

It's not pretty to have loop variables exposed. But this seems better than having an `r_target.push_back` which could trigger vector reallocations.

inst/include/IndividualIndex.h
giovannic: The data structure, either 32 or 64 bit, should work for an unlimited number of individuals (theoretically). Each uint can store the state of 32/64 individuals and we store a vector to get our population size.

Templating to allow 32 would only open up optimisations for 32 bit machines.

inst/include/IndividualIndex.h
giovannic: hahah great catch. I was expecting `ctzll(0ULL) == 64`. I will add a check

giovannic:
> OK this is pretty dense, but I think I can get about there.

Thanks for getting through it all!

> What's going on with the massive vector realloc insert box? That's come down ~10% from the total (looking at `apply_state_update`) but is still pretty big? This is doing an insert into a vector (`vec[i] = x` or similar?). Is this unavoidable without further large refactors? I might be totally up the wrong tree here sorry.

It's a very interesting question. Yes, this ends up assigning to a vector, something like `bitset[position] = true`. But the bitset should already be allocated! And vector accesses with `[]` do not allocate new memory, they have undefined behaviour for out of bounds. Which makes this very weird.

I want to to investigate it further but it feels like it's drifting away from #9. Let's see how much time I have for it when I pass through all these comments :)

inst/include/IndividualIndex.h
giovannic: I have templated this anyway. For every new type we want to support, we just need to overload another (ctz)[https://github.com/mrc-ide/individual/pull/13/files#diff-54d50fea500dd4ef4d4d335c32b2278aR89]

inst/include/Scheduler.h
giovannic: The PR diff won't update :/ [github issue](https://github.com/isaacs/github/issues/750)

But the documentation is there: https://github.com/mrc-ide/individual/blob/feat/ctzl_linear/inst/include/Scheduler.h

PR #14, opened 2020-07-02
Allow clients to render vectors

giovannic:


giovannic:
Don't like this feature

PR #24, opened 2020-08-20
Feat/ignore fill

### LINKED ISSUE #21: Unexpected variable fill
I know this is [documented as a feature](https://mrc-ide.github.io/individual/reference/SimAPI.html#method-queue_variable_update), but it's highly prone to error.

The expected case is great:

```r
# I want to reset everyone's immunity to 0:
api$queue_variable_update(human, immunity, 0)
# Everyone is set to 0
```

But it leads to unexpected behaviour if I give an index:

```r
# I want to reset a random subset of human immunities to 0
n_died <- rbinom(10, .5)
died <- sample.int(10, n_died, replace=FALSE)
api$queue_variable_update(human, immunity, 0, died)
# If n_died == 0 everyone will be reset to 0!
# I really want no one to be set to 0
```

I propose a change to the R and C++ api to make variable updates with empty indices do nothing.

pwinskill:
Currently failing CI on Travis, but I can't understand the error - maybe you coud take a look @giovannic and see if it means something to you?

Should any of the format tidying and/or documentation corrections be included here before the merge?

Closes #21


R/api.R
```diff
@@ -65,66 +65,85 @@ SimAPI <- R6::R6Class(
...
-    #' @description Create an R wrapper for the API
-    #' @param cpp_api the cpp implementation of the simulation api
-    #' @param parameters model parameters
-    #' @param renderer renderer to store model outputs to
-    initialize = function(cpp_api, parameters, renderer) {
-      private$.api <- cpp_api
-      private$.parameters <- parameters
-      private$.renderer <- renderer
-    }
+
```
giovannic, line 96: The indenting here is a bit weird, which is causing the diff.

I _think_ if you remove one space from these lines it should remove the diff

giovannic (changes requested):
Sorry for the delay! Completely missed it.

I've push a change for travis. Hopefully that fixes the build

PR #27, opened 2020-09-25
Remove Rcpp from state creation

giovannic:
This PR simplifies creating variables. It has breaking changes for old models.

Before you'd have to write:

```r
immunity <- Variable$new('immunity', function(size) runif(size, 0, .2))
```
And the second argument would be run to initialise the immunity variable.

But now you would write:

```r
immunity <- Variable$new('immunity', runif(size, 0, .2))
```
Passing in a vector of initial values. The package will later validate that the vector is the correct size.


R/individual.R
```diff
@@ -35,8 +38,16 @@ Individual <- R6::R6Class(
         stop('No duplicate event names allowed')
       }
 
+      population_size <- sum(vnapply(states, function(s) s$initial_size))
+
+      variable_sizes <- vnapply(variables, function(v) length(v$initial_values))
+      if (any(variable_sizes != population_size)) {
+        stop("variable initial values must match the population size")
```
pwinskill, line 45: Assumming all variables will be named, it would be helpful to the user here to specify in the error message which variable caused this error if there's not too much overhead in doing so.

tests/testthat/test-updates.R
```diff
@@ -18,8 +19,9 @@ test_that("updating variables works", {
 })
 
 test_that("updating variables at the boundaries works", {
```
pwinskill, line 21: Should a test at the boundaries include the first index as well as the last?

tests/testthat/test-updates.R
```diff
@@ -50,8 +53,9 @@ test_that("updating variables with an empty index is ignored", {
 })
 
 test_that("updating variables with silly indecies errors gracefully", {
```
pwinskill, line 55: As well as this test. Is it worth ensuring a graceful error if the user mis-specifies the wrong type, e.g. if I tried: `sim$r_api$queue_variable_update(human, sequence, "A", 10)`
Maybe this is cheked elsewhere

pwinskill, line 55: typo indecies

pwinskill (commented):
Hey @giovannic 
Looks like a helpful change that I think will feel more comfortable to an R user. A few minor comments/suggestions.

giovannic:
Thanks! Made the changes

Issue #40, opened 2020-11-09
Prefab for forks in transitions


OJWatson:
A common IBM transition I can see being really useful would be for transitions where individuals transition at a rate (fixed or different per individual) into one of two states determined by an additional probability (maybe templating fixed or different per individual would be great). 

```
Rcpp::XPtr<process_t> fixed_probability_forked_state_change_process(
    const std::string individual,
    const std::string from_state,
    const std::string to_state_1,
    const std::string to_state_2,
    double rate,
    double fork_probability) { 

    // Draw which individual will transition based on rate

    // Allocate drawn individuals to either to_state_1 or to_state_2 based on for_probability

}
```

pwinskill:
It might be possible to generalise this into one of N states, given a rate and vector of probabilities?

giovannic:
🤓 Not a bad idea at all!

```cpp
Rcpp::XPtr<process_t> fixed_probability_forked_state_change_process(
    const std::string individual,
    const std::string source_state,
    const std::vector<std::string> destination_states,
    double rate,
    const std::vector<double> destination_probabilities) { 

    // Draw which individuals will transition based on `rate`

    // Distribute those individuals to the distination states based on their corresponding probabilities

}
```

slwu89:
@giovannic is this good to close? I think #63 covers it.

PR #45, opened 2021-02-06
add prefab for forked transitions

slwu89:
Implement a feature to address issue #40 , adds a new function `fixed_probability_forked_state_change_process` taking the same signature as recommended by @giovannic, also added a new test in test-prefab.R to check the function is working correctly.


src/prefab.cpp
```diff
@@ -76,6 +76,51 @@ Rcpp::XPtr<process_t> fixed_probability_state_change_process(
...
+            // sample if each person leaves; if so sample their destination
+            for (const auto individual : target_individuals) {
+                if (random[random_index] < rate) {        
+                    double u = R::runif(0.,1.);           
+                    auto k_it = std::upper_bound(cdf.begin(),cdf.end(),u);
+                    int k = std::distance(cdf.begin(),k_it);
+                    individuals_to_dest[k].emplace_back(individual);
+                }
+                ++random_index;
+            }
```
giovannic, line 114: We could save on calling R.runif (which is quite expensive) in a loop if we have something like:

```suggestion
            std::vector<size_t> source_individuals(
                target_individuals.begin(),
                target_individuals.end()
            ); //perhaps some less confusing variable names here
            
            // get the indices of source_individuals who leave with only two calls to RNG
            auto leaving_individuals = Rcpp::sample(
              source_individuals.size(),
              Rcpp::rbinom(1, source_individuals.size(), rate)[0], //sample the number of individuals leaving
              false, // replacement
              R_NilValue, // evenly distributed
              false // one based
            )
            
            // then use your lovely CDF
            for (const auto i : leaving_individuals) {        
                double u = random[random_index];           
                auto k_it = std::upper_bound(cdf.begin(),cdf.end(),u);
                int k = std::distance(cdf.begin(),k_it);
                individuals_to_dest[k].emplace_back(source_individuals[i]);
                ++random_index;
            }
```

tests/testthat/test-prefab.R
```diff
@@ -20,6 +20,56 @@ test_that("fixed_probability_state_change moves a sane number of individuals aro
   expect_equal(n_s + n_i, 11)
 })
 
+test_that("fixed_probability_forked_state_change_process works properly", {
```
giovannic, line 23: Thank you, this test is very concise and clear

tests/testthat/test-prefab.R
```diff
@@ -20,6 +20,56 @@ test_that("fixed_probability_state_change moves a sane number of individuals aro
...
+  n <- 2e5
+  A <- State$new('A', n)
+  B <- State$new('B', 0)
+  C <- State$new('C', 0)
+  D <- State$new('D', 0)
+  E <- State$new('E', 0)
+  F <- State$new('F', 0)
```
giovannic, line 30: Could we avoid overriding F, which is short for FALSE?

giovannic (changes requested):
Thank you, this is great. I've suggested a few changes inline

src/prefab.cpp
slwu89: Thanks for the comments! I did not realize `R::runif` was such a performance hit.  I have reworked the function based on your suggestions here https://github.com/mrc-ide/individual/blob/a192e5c336eccaebc3c4596febe91cdc69a87f3d/src/prefab.cpp#L98 (sorry, not sure how to get the nice in-comment code display like you have!)

tests/testthat/test-prefab.R
slwu89: Good point, I have changed it so it no longer overrides any of R's objects.

src/prefab.cpp
giovannic: Those fancy in-comment displays are "code suggestions"

>\`\`\`suggestion
> put your code here
> \`\`\`

Multiline suggestions are [brand new](https://github.blog/changelog/2019-10-01-multi-line-references-and-comments-beta/) ;)

Yesss, random number generation looks innocuous but it can [bite back](https://blog.revolutionanalytics.com/2012/08/benchmarking-random-number-generation-from-c.html)

src/prefab.cpp
```diff
@@ -76,6 +76,57 @@ Rcpp::XPtr<process_t> fixed_probability_state_change_process(
...
+                false // one-indexed
+            );
+            // indicies of individuals going to each destination state
+            std::vector<std::vector<size_t>> individuals_to_dest(n_dest);
+            const auto& random = Rcpp::runif(n_leave);
+            auto random_index = 0;
+            for (const auto individual : leaving_individuals) {
+                auto dest_iter = std::upper_bound(cdf.begin(), cdf.end(), random[random_index]);
+                int dest = std::distance(cdf.begin(), dest_iter);
+                individuals_to_dest[dest].emplace_back(individual);
```
giovannic, line 118: `leaving_individuals` is an integer between 0 and `target_individuals.size()`

What we really want to add to `individuals_to_dest[dest]` is someone from `target_individuals` at position `individual`

This could move someone to `dest` who isn't in `target_individual`

giovannic:
One small change and we can merge :)

src/prefab.cpp
slwu89: Thanks @giovannic, yes I overlooked that the index of the bitset that holds people's unique integer id is not always the same as that integer id! That would not be a fun bug to track down later on. I've gone back to the design you originally suggested, copying the bitset elements into a `std::vector<size_t>` object , and then passing the sampled elements of that to `queue_state_update` so that everyone getting moved really was in the `target_individual` object.

src/prefab.cpp
slwu89: Interesting! I thought that `R::runif` would be super fast since it was going down to R's C API. If I wanted to draw the random numbers one by one in a fast way in the future, would I include Rmath.h and use `unif_rand`?

slwu89:
Thanks for the comments @giovannic! I hope this commit (https://github.com/mrc-ide/individual/pull/45/commits/8af7d00a18fbe9a7e65d2767561d0af36aeda026) fixes that problem of mixing up the index with the actual integer id of the individual in the bitset. 

I had a thought to avoid making the other vector `source_individuals` into which the elements of the bitset are copied, but I don't know enough about the iterators of the IterableBitset object to know if this is more or less efficient (I guess I can just code both and profile it). My thoughts are this:

1. sort the uniformly sampled indices in `leaving_individuals` in ascending order
2. grab an iterator to the beginning of `target_individuals`
3. use `std::advance` with the sorted vector of indices to hop along the `target_individuals` bitset, adding elements to `queue_state_update` as we go.

This would save the creation/destruction of the vector `source_individuals` but I'm not sure if that's a big concern, overall.

giovannic (approved):

giovannic:
Yes! Roughly, the vector conversion would be O(bitset.max_size) and sorting would be O(size*log2(size)). But we would need to profile see the improvement in practice.

Issue #49, opened 2021-02-17
Bitset to vector  


slwu89:
Hi! I have a question about how to properly use the Bitset class, in particular the `to_vector` member function seems to be giving odd results for me. An example of unexpected behavior is here.

```
> new <- Bitset$new(4)
> new$insert(15:18)
> new$to_vector()
[1] 0 0 0 0
```
What is the proper way to initialize a bitset when the number of desired integers to insert is known? Thanks!

giovannic:
hmm,

We need some error checking here. Something like:

```r
b <- Bitset$new(size = 4)
b$to_vector()
# > numeric(0)
b$insert(15:18)
# > Error: You cannot insert elements out of range
b$insert(c(2, 4))$to_vector()
# > c(2, 4)
```

Currently, if you want to create a bitset for known values, you'd do something like:

```r
b <- Bitset$new(size = 30)$insert(15:18)
```

Perhaps something like this would be nicer?

```r
b <- Bitset$new(size = 30, values = 15:18)
```

giovannic:
Actually, the implementation already exists, would simply change [this](https://github.com/mrc-ide/individual/blob/release/0.1.0/src/bitset.cpp#L32) to `b->insert_safe`

slwu89:
Ah, the "problem" in that case is that I just don't understand the relationship between the `size` argument in the bitset's constructor and how many integer objects it can store. Maybe a clearer way to ask the question is this, given a vector, say `std::vector<int> vec = {1,2,3,4,5}`, what would the `size` argument for the IterableBitset constructor be if I wanted to create one from `vec`? Like `individual_index_t bitset(size, vec.begin(), vec.end())`. I was getting some errors when using `vec.size()` for that argument.

giovannic:
Ah! So IterableBitset stores a set of integers in [0, n), as is expected in C++, and Bitset stores a set of integers in [1, n], as is expected in R.

So if you're making a bitset to store `std::vector<size_t> vec = {1,2,3,4,5}`. You'd have to make size >= 6, because it expects 0 as well!

However `Bitest$new(size = 5)$insert(1:5)` would work fine. Because under the hood it stores `(1:5) - 1`.

slwu89:
That's very clear! Thanks. I was mentally using the IterableBitset like a std::vector<int> object, when I should have been thinking like std::set<int>. I think this issue can be closed now if you are happy with this resolution (I am).

PR #50, opened 2021-02-17
Add range checking for bitset inserts

giovannic:
Partially resolving #49


slwu89 (approved):
Looks great! Nothing to add here.

PR #54, opened 2021-02-18
Make event time start at 1

giovannic:
Event listeners can be triggered with a timestep of 0. This fixes it

### DIFF
--- a/inst/include/Event.h
+++ b/inst/include/Event.h
@@ -29,7 +29,7 @@ inline std::vector<size_t> round_delay(const std::vector<double>& delay) {
 }
 
 struct EventBase {
-    size_t t = 0;
+    size_t t = 1;
 
     virtual void tick() {
         ++t;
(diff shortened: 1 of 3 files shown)


R/render.R
```diff
@@ -18,7 +18,7 @@ Render <- R6::R6Class(
     #' scalar outputs to store to the final render
     initialize = function(timesteps) {
       private$.timesteps = timesteps
-      private$.vectors[['timestep']] <- seq_len(timesteps)
+      private$.vectors[['timestep']] <- seq(timesteps)
```
slwu89, line 21: I would have thought `seq_len` would have given you exactly what you want, I think `seq` relies on it parsing the arguments in some specific way?

slwu89 (commented):
Looks great! (of course).

R/render.R
giovannic: true! seq_len is more specific

PR #56, opened 2021-02-19
implement an IntegerVariable class (feat/int_var)

slwu89:
This PR implements an `IntegerVariable` class that largely resembles the `DoubleVariable` class but has a new method, `get_index_of` that returns the indices (as a `Bitset`) of individuals whose values are either in a given set or with a range [a,b]. The PR also includes passing tests to confirm the new class works properly.


R/integer_variable.R
```diff
@@ -0,0 +1,99 @@
...
+    },
+
+
+    #' @description return a bitset for individuals with some subset of values
+    #' Either search for indices corresponding to values in \code{set}, or
+    #' for indices corresponding to values in range [a,b].
+    #' @param set a vector of values 
+    #' @param a lower bound
+    #' @param b upper bound
+    get_index_of = function(set = NULL, a = NULL, b = NULL) {
```
giovannic, line 40: :o I love this feature

R/integer_variable.R
```diff
@@ -0,0 +1,99 @@
...
+      } else if(is.numeric(index)) {
+        if (length(index) != 0) {
+          integer_variable_queue_update(
+            self$.variable,
+            values,
+            index
+          )
+        }
+      } else if(!is.null(index$.bitset)) {
+        integer_variable_queue_update(
```
giovannic, line 89: Ah, this causes a bug. I've just fixed it in #55 for DoubleVariable. Could you make a similar update here?

inst/include/IntegerVariable.h
```diff
@@ -0,0 +1,148 @@
...
+        
+        std::vector<size_t> result_ix;
+        for(size_t it = 0; it < values.size(); it++) {
+            if( !(values[it] < a) && !(b < values[it]) ) {
+                result_ix.push_back(it);
+            }
+        }
+
+        if (!result_ix.empty()) {
+            size_t bset_size = *std::max_element(result_ix.begin(), result_ix.end());
```
giovannic, line 80: Interesting, the idea here is that we could save space if the indices are small? Kinda nice.

But the issue is that [bitmap bitwise operations require the same sized bitmaps](https://github.com/mrc-ide/individual/blob/dev/inst/include/IterableBitset.h#L239)

And it would be really nice to write something like:

```r
is_infected <- CategoricalVariable$new(...)
risk_level <- IntegerVariable$new(...)
target <- is_infected$get_index_of('yes')$and(risk_level$get_index_of(a=3, b=5))
```

Perhaps it's best to allocate a bitset for all individuals and set the values directly?

giovannic (changes requested):
This is really useful! I can't wait. A few small requests

R/integer_variable.R
slwu89: wonderful! It's something I'll need in my future uses for individual.

inst/include/IntegerVariable.h
slwu89: Okay, got it. I think I _finally_ understand how bitset works. I've gone ahead and made a change that I think should address this issue: https://github.com/mrc-ide/individual/pull/56/commits/ad8ab82d8bab3544d961fd656a416798198efe69

The returned bitset is initialized with a size equal to that of the number of values held in the IntegerVariable itself.

R/integer_variable.R
slwu89: done! https://github.com/mrc-ide/individual/pull/56/commits/d0ee210550663da2eb1eca7b0b0e1d77bdffea2e# I've confirmed that it is fixed.

inst/include/IntegerVariable.h
slwu89: And just confirmed that your example does indeed work correctly!

giovannic (approved):
Great!

PR #57, opened 2021-02-23
query number of individuals in set for all variable types

slwu89:
This PR adds `get_size_of` methods to IntegerVariable, CategoricalVariable, and DoubleVariable to query the number of individuals in a state without returning a bitset. For IntergerVariable  the user may either ask for the number of values in some set of value or in a interval [a,b]. For DoubleVariable the user asks for the number of values within some interval [a,b]. CategoricalVariable takes a vector of strings as the set to be matched against.

This also adds `get_index_of` to DoubleVariable to return a bitset of individuals whose value is within the interval [a,b].

Please let me know if I am submitting this PR into the wrong upstream branch! I think dev is the right one for it to go but I'm not certain.

### DIFF
--- a/R/RcppExports.R
+++ b/R/RcppExports.R
@@ -57,6 +57,10 @@ categorical_variable_get_index_of <- function(variable, values) {
     .Call(`_individual_categorical_variable_get_index_of`, variable, values)
 }
 
+categorical_variable_get_size_of <- function(variable, values) {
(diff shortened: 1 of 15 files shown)


giovannic (approved):
Thanks! This could bring some minor speed improvements.

Could you add yourself to the contributors' list in DESCRIPTION? You could push that in a seperate PR if you like

PR #58, opened 2021-02-23
Feat/filter bitset

giovannic:
Adds behaviour for users to filter in the i-th elements of a bitset.

Let me know if the interface makes sense.


R/bitset.R
```diff
@@ -73,3 +73,19 @@ Bitset <- R6::R6Class(
...
   )
 )
+
+#' @title Filter a bitset
+#' @description filter in the i-th values in the set (assuming the bitset values
+#' are in ascending order), for all i in other
+#' @param bitset the bitset to filter
+#' @param other the values of bitset to keep
+#' @export
+filter_bitset = function(bitset, other) {
```
slwu89, line 83: This is a cool addition! I think the docs can be improved a bit for general users. I might suggest something like:

```
@description This non-modifying function returns a new \code{\link{Bitset}} object of the same maximum size as the original but which only contains those values at the indices specified by the argument `other`.
Indices in `other` may be specified either as a vector of integers or as another bitset. 
Please note that filtering by another bitset is not a "bitwise and" intersection, and will have the same behavior as providing an equivalent vector of integer indices.
```

slwu89 (changes requested):
This is a neat feature! I just have a suggestion regarding the documentation for `filter_bitset` to clarify some points about what it does.

PR #62, opened 2021-02-24
Remove get_scheduled from the Event interface

### LINKED ISSUE #61: Bug in Event$get_scheduled
The `get_scheduled` method in Event calls `event_get_scheduled` which wants an XPtr to a TargetedEvent but gets an Event passed to it, causing a crash when it tries to access things that don't exist.

giovannic:
Closes #61


slwu89 (approved):
Cool. Thanks!

Issue #39, opened 2020-11-09
Prefab for different transition rates for each individual


OJWatson:
We have a prefab for fixed transition rates:

```
Rcpp::XPtr<process_t> fixed_probability_state_change_process(
    const std::string individual,
    const std::string from_state,
    const std::string to_state,
    double rate) { ...
```

Would be really helpful to have a very similar one but for different rates for each individual:

```
Rcpp::XPtr<process_t> multi_probability_state_change_process(
    const std::string individual,
    const std::string from_state,
    const std::string to_state,
    std::vector<double> rate) { ...
```

slwu89:
@giovannic is this one good to close too? PR #63

PR #63, opened 2021-02-25
additional prefab transition processes

slwu89:
This PR adds functions `fixed_probability_multinomial_process`, `multi_probability_multinomial_process`, and `multi_probability_bernoulli_process` which address issues #40 and #39 , allowing for multinomial processes with constant and individual-specific probability of leaving the source compartment, and Bernoulli process with individual specific probability. The individual-specific probabilities are expected to be contained in a `DoubleVariable` object.

This PR also moves the logic in `bitset_sample` (in bitset.cpp) to `bitset_sample_internal` in IterableBitset.h; this was needed because the prefab transitions in prefab.cpp needed to use that function's logic. It also has a new function `bitset_sample_multi_internal` in IterableBitset.h which samples a bitset with templated input iterators, necessary for `multi_probability_multinomial_process` to work correctly.

Tests for the prefab functions have also been added!


src/prefab.cpp
```diff
@@ -0,0 +1,182 @@
...
+//' probabilities contained in the vector 'destination_probabilities'.
+//' @param variable a \code{\link{CategoricalVariable}} object
+//' @param source_state a string representing the source state
+//' @param destination_states a vector of strings representing the destination states
+//' @param rate probability of individuals in source state to leave
+//' @param destination_probabilities probability vector of destination states
+//' @export
+// [[Rcpp::export]]
+Rcpp::XPtr<process_t> fixed_probability_multinomial_process(
+    const Rcpp::Environment variable,
```
giovannic, line 27: Could we instead pass an Rcpp::XPtr<DoubleVariable>?

A few reasons:
 * It adds a bit of type checking
 * I have found incomprehensible errors when I accidentaly put in an R thing that is not what we expect and gets casted into an Rcpp::Environment

Then we could have a small function in R/prefab.R which checks the CategoricalVariable and passes .variable to this function

src/prefab.cpp
```diff
@@ -0,0 +1,182 @@
...
+//' probabilities contained in the vector 'destination_probabilities'.
+//' @param variable a \code{\link{CategoricalVariable}} object
+//' @param source_state a string representing the source state
+//' @param destination_states a vector of strings representing the destination states
+//' @param rate_variable \code{\link{DoubleVariable}} giving individual probability of each individual in source state to leave
+//' @param destination_probabilities probability vector of destination states
+//' @export
+// [[Rcpp::export]]
+Rcpp::XPtr<process_t> multi_probability_multinomial_process(
+    const Rcpp::Environment variable,
```
giovannic, line 90: Same as above

src/prefab.cpp
```diff
@@ -0,0 +1,182 @@
...
+//' individual probability specified by the \code{\link{DoubleVariable}}
+//' object 'rate_variable'.
+//' @param variable a \code{\link{CategoricalVariable}} object
+//' @param from a string representing the source state
+//' @param to a string representing the destination state
+//' @param rate_variable \code{\link{DoubleVariable}} giving individual probability of each individual in source state to leave
+//' @export
+// [[Rcpp::export]]
+Rcpp::XPtr<process_t> multi_probability_bernoulli_process(
+    const Rcpp::Environment variable,
```
giovannic, line 156: same as above

giovannic (changes requested):
Very nice! Few small nit picks

src/prefab.cpp
slwu89: Totally. I wasn't sure the best way to pass those data down to C++, this is much better than using Rcpp::Environment.

Fixed in https://github.com/mrc-ide/individual/pull/63/commits/a3cba8683426608a9d8d1969cd68456157b75ac5 and also made `bitset_sample_multi_internal` a little simpler.

giovannic (approved):
Great, thanks!

PR #64, opened 2021-02-25
Fix bug where Bitset$not produced ghost bits

giovannic:


slwu89 (approved):
awesome, no user facing changes and the bitset operations are still somewhat murky to me, so it all looks good from my perspective!

PR #65, opened 2021-03-02
bugfix for C++ event listeners

slwu89:
The function Event$.process calls a nonexistent function `process_listener` to handle listeners that are XPtr objects to the `listener_t` type. This PR defines that function `process_listener` and one for targeted listeners `process_targeted_listener` in src/event.cpp. It also defines a new method for Event and TargetedEvent class `.process_listener_cpp` which is used in `.process` when a XPtr listener is encountered, otherwise inheritance in TargetedEvent will work incorrectly, as the `targeted_listener_t` type needs the second argument for the bitset of targeted individuals.

The codecov is negative but we don't have any existing tests of event listeners implemented in C++ to base off.


giovannic (approved):
Great catch. Looks good to me

PR #68, opened 2021-03-05
detailed API vignette & various documentation clarifications 

slwu89:
This PR does 2 things:

1. Adds a vignette API.Rmd that hopefully explains in quite a bit of detail the objects exposed to users at the R level that they can use to make their individual based simulations, and how they interact. 
2. Makes various clarifications throughout the roxygen2 docs for objects, methods, etc. In particular I tried to add links to other internal help files when appropriate, and clarify the types of allowed inputs for function arguments. In a few cases I modified the R function code a bit to do type checking against Bitset arguments, and if not use `as.integer` on the input vector (which AFAIK will be stored as `numeric` in R by default).

I haven't gone through and done the SIR tutorial vignette or the vignette that discusses writing custom C++ to speed up certain listeners or processes, but I think enough has been done that a review of the current work would be helpful!

### DIFF
--- a/vignettes/API.Rmd
+++ b/vignettes/API.Rmd
@@ -25,7 +25,11 @@ library(individual)
       1. [Categorical Variables](#cat_var)
       2. [Integer Variables](#int_var)
       3. [Double Variables](#dbl_var)
-  5. [References](#ref)
+  3. [Processes](#proc)
+      1. [Render](#rend)
+  4. [Events](#event)
+      1. [Targeted Events]($t_event)
+  5. [Simulate](#sim)
 
 ## Introduction {#intro}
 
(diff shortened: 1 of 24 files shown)


vignettes/API.Rmd
```diff
@@ -0,0 +1,168 @@
...
+      variable$get_index_of(from)$sample(rate)
+    )
+  }
+}
+```
+
+We can test empirically that dwell times do indeed follow a Geometric distribution (and demonstrate uses of `IntegerVariable` objects) by modifying slightly the function to take `int_variable`, an object of class `IntegerVariable`. Each time the returned function closure is called it randomly samples a subset of individuals to move to the destination state, and records the time step at which those individuals move in the `IntegerVariable`. We run the simple "simulation" loop until there are no more individuals left in the source state, and compare the dwell times in the source state to the results of `rgeom`, and see that they are from the same distribution.
+
+```{r}
+bernoulli_process_time <- function(variable, int_variable, from, to, rate) {
```
giovannic, line 82: A very elegant way of explaining the bernoulli process!

I'm excited and scared about documenting custom simulation loops this early on :D

giovannic (approved):
One comment. Thanks for the fixes too!

vignettes/API.Rmd
slwu89: Ah, yeah after the other vignettes are done I'll make it as clear as possible this vignette should only be read after the tutorial ones! But it's nice to show off a bit just how flexible the software here is!

PR #72, opened 2021-03-12
prefab for age-structured transmission

slwu89:
This PR adds a new prefab `infection_age_process` to handle age structured contact matrices, where each agent has an age class (IntegerVariable) and a social mixing/contact matrix of the type from [Prem et al (2017)](https://journals.plos.org/ploscompbiol/article?id=10.1371/journal.pcbi.1005697) or [POLYMOD/socialmixr](https://cran.r-project.org/web/packages/socialmixr/). The C++ implementation is sufficiently complex and the corresponding R version sufficiently slow that I think it would be worth including in the package (this specific use case arose from my own project).

This PR also simplifies the `get_index_of` and `get_size_of` methods in the variables to not make a temporary `vector<int>` and instead insert into the returned bitset directly, which should be a little bit faster. It also adds overloads for when the desired categories/values are single `string` or `int` rather than vectors, which is more convenient when using them in C++.

Unfortunately the overloads for single values rather than vectors decrease test coverage (I don't know a good way to export them to R because the names will clash, and it seems silly to have a check at the R level if something is length 1 or >1,  the minor speed gain from using one v. the other is probably offset by the conditional statement in R...).


inst/include/IntegerVariable.h
```diff
@@ -45,34 +45,46 @@ struct IntegerVariable : public Variable {
...
             if(findit != values_set.end()){
-                result_ix.push_back(it);
+                result.insert(it);
             }
         }
 
+        return result;
+    }
+
+    // get indices of individuals whose value is in some set
```
giovannic, line 60:
```suggestion
    // get indices of individuals whose value is in a singleton set
```

inst/include/IntegerVariable.h
```diff
@@ -45,34 +45,46 @@ struct IntegerVariable : public Variable {
...
+    virtual individual_index_t get_index_of_set(
+        const int values_set
```
giovannic, line 62:
```suggestion
        const int value
```

inst/include/IntegerVariable.h
```diff
@@ -45,34 +45,46 @@ struct IntegerVariable : public Variable {
...
+    ) const {
+
         auto result = individual_index_t(size);
-        result.insert_safe(result_ix.begin(), result_ix.end());
+        for(size_t it = 0; it < values.size(); it++){
+            if (values[it] == values_set) {
```
giovannic, line 67:
```suggestion
            if (values[it] == value) {
```

giovannic (changes requested):
V. nice.

I like the removal of result_idx.

We're soon going to have to find a way to organise the prefabs.

Just a minor comment.

slwu89:
> V. nice.
> 
> I like the removal of result_idx.
> 
> We're soon going to have to find a way to organise the prefabs.
> 
> Just a minor comment.
> 
> [edit]
> Also to keep the coverage up, we'd need tests getting the index for unknown categories

I might be misunderstanding your comment but doesn't this test in test-variables.R handle that already? 

```
test_that("getting a non registered state index fails", {
  state <- CategoricalVariable$new(
    c('S', 'I'),
    c(rep('S', 10), rep('I', 100))
  )

  expect_error(
    state$get_index_of('R'),
    '*'
  )
})
```

Agree on the prefabs...perhaps we can stick with this set of prefabs for CRAN submission and then figure out what to do? We could do better in-package management, maybe with naming conventions `process_something_something`, `render_something_something` and multiple .cpp files, or if it looks like there will be _a lot_ of new prefabs being made perhaps break it into another package? But that does lose some of the elegance...

giovannic:
>  test-variables.R handle that already?

That's so strange! I'll have to look into the how covr works. Because it was giving me this output...
 
Screenshot 2021-03-12 at 17 02 07

We'll ignore it for now.

> perhaps we can stick with this set of prefabs for CRAN submission

Yup, this is definitely something for later

giovannic (approved):

slwu89:
> > test-variables.R handle that already?
> 
> That's so strange! I'll have to look into the how covr works. Because it was giving me this output...
> 
> Screenshot 2021-03-12 at 17 02 07
> 
> We'll ignore it for now.
> 
> > perhaps we can stick with this set of prefabs for CRAN submission
> 
> Yup, this is definitely something for later

Ah, that test is covering the `get_index_of` that takes the `vector<string>` as its argument rather than just `string`... I guess that is why. Yeah I don't know a good way to do the tests of the overloads.

giovannic:
> Yeah I don't know a good way to do the tests of the overloads.

Ahh ok, yes some tests will have to be C++. I'm still figuring out how to make them look nice, and they do feel intimidating at first.

For the future, something like this should do the trick:

```cpp
// in test-categorical_variable.cpp
#include <Rcpp.h>
#include <testthat.h>

context("CategoricalVariable") {

          test_that("CategoricalVariable can return index of category") {
                const auto x = CategoricalVariable({"S", "I", "R"}, {"I", "I", "R"});
                const auto expected = individual_index_t(3, {0, 1});
                const auto actual = x.get_index_of("I");
                expect_true(expected == actual);
        }

        test_that("CategoricalVariable errors for an unknown category category") {
                const auto x = CategoricalVariable({"S", "I", "R"}, {"I", "I", "R"});
                expect_error(x.get_index_of("Z"));
        }

}
```

PR #76, opened 2021-03-14
Add xor for two bitsets

slwu89:
Being able to do XOR on two bitsets is useful when dealing with complicated divisions of state space, this PR adds (and tests) that functionality.

### DIFF
--- a/src/bitset.cpp
+++ b/src/bitset.cpp
@@ -80,6 +80,14 @@ void bitset_xor(
     const Rcpp::XPtr<individual_index_t> a,
     const Rcpp::XPtr<individual_index_t> b
     ) {
+    (*a) ^= (*b);
+}
+
+//[[Rcpp::export]]
+void bitset_set_difference(
+    const Rcpp::XPtr<individual_index_t> a,
+    const Rcpp::XPtr<individual_index_t> b
+    ) {
     (*a) &= ~(*b);
 }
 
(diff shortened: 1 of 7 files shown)


src/bitset.cpp
```diff
@@ -75,6 +75,14 @@ void bitset_or(
     (*a) |= (*b);
 }
 
+//[[Rcpp::export]]
+void bitset_xor(
+    const Rcpp::XPtr<individual_index_t> a,
+    const Rcpp::XPtr<individual_index_t> b
+    ) {
+    (*a) &= ~(*b);
```
giovannic, line 83: This appears to implement

```r
a$and(b$not())
```

Which _is_ set difference, but is not _xor_.

Which do you want to implement? Both isn't a bad idea.

To implement xor I would do 
```suggestion
    (*a) ^= *b;
```

To do set difference, I'd add something like this to the Bitset R6 object:

```r
difference = function(other) self$and(other$not())
```

giovannic (changes requested):

src/bitset.cpp
slwu89: Oh interesting! Yeah I think both is a good idea. I'll work that one out.

slwu89: "xor" properly renamed "set_difference", and proper "xor" added. Included more significant tests for both, for disjoint, identical, and partially overlapping sets.

giovannic (approved):

PR #82, opened 2021-03-18
Fix bug in API docs:

giovannic:
* Change TargetedEvent to take population size
* Remove same bug from the README and remove redundant example code


slwu89:
nice catch!

slwu89 (commented):
nice catch!

PR #83, opened 2021-03-24
Feat/performance vignette

slwu89:
This PR adds a bit more detail to the Bitset section of the Performance vignette, in particular the 2 new methods, and explains more about initialization of a bitset object.


giovannic (commented):
Fantastic.

Could we also mention that $to_vector is slow?

For large variables, if you want to extract the value of each individual *more* than you want to perform set operations, a vector based variable (double/integer) is more appropriate.

giovannic (approved):

PR #86, opened 2021-03-29
Feature/default render

pwinskill:
Adding functionality to `Render` to allow the user to set the default value for a variable.

This is useful if the user will render variables that won't get called every timestep, and will want to set  a default, instead of NA (most likely to 0).

There seems to be a few different ways this feature could be implemented - this one seems fairly tidy.


giovannic (approved):
This is much better than I imagined doing it.

PR #90, opened 2021-05-01
revise docs for remaining errors or unclear language

slwu89:
this PR clears up remaining minor issues with the documentation and vignettes for clarity. Also, the simulation loop graphic from the JOSS manuscript is now included in the API vignette.


vignettes/API.Rmd
```diff
@@ -49,15 +49,15 @@ There are 3 types of variable objects: `individual::CategoricalVariable` for dis
...
+Most epidemiological models will *require* use of categorical variables to define state, for example, the Susceptible, Infectious, and Recovered classes in the classic [SIR model](https://en.wikipedia.org/wiki/Compartmental_models_in_epidemiology#The_SIR_model). In general, for compartmental models, each set of compartments which describes an aspect of an individual should be mapped to a single `individual::CategoricalVariable`. For example, an SIR model which also classified individuals as belonging to high and low risk groups could be represented with two categorical variable objects.
```
giovannic, line 52:
```suggestion
Most epidemiological models will *require* use of categorical variables to define state, for example, the Susceptible, Infectious, and Recovered classes in the classic [SIR model](https://en.wikipedia.org/wiki/Compartmental_models_in_epidemiology#The_SIR_model). In general, for compartmental models, each set of compartments which describes an aspect of an individual should be mapped to a single `individual::CategoricalVariable`. For example, a model which classifies individuals as in an SIR state _and_ in a high or low risk groups could be represented with two categorical variable objects
```

giovannic (approved):

PR #92, opened 2021-05-13
add method to Bitset to check existence of elements

slwu89:
For some of the causal estimation work, it was useful to be able to pass a integer vector of elements to a bitset and check if their bits were set, returning a logical vector. This PR adds a method `IterableBitset::exists_safe` which calls `IterableBitset::exists` if the requested element is in range, a function `bitset_sample_vector` in bitset.cpp to be called from R, and a method to the R6 object. A simple test is included.

I'm not sure why some number of man/*.Rd files are marked as being changed, but they should be referencing the most recent roxygen documentation from the dev branch.


src/utils.h
```diff
@@ -16,4 +16,10 @@ inline void decrement(std::vector<size_t>& x) {
     }
 }
 
+inline void decrement(Rcpp::IntegerVector& x) {
+    for (auto i = 0u; i < x.size(); ++i) {
+        --x[i];
+    }
+}
+
```
giovannic, line 24: Don't we have this somewhere already?

inst/include/IterableBitset.h
```diff
@@ -390,6 +391,16 @@ inline void IterableBitset<A>::insert_safe(size_t v) {
     insert(v);
 }
 
+//' @title exists_safe
+//' @description check if an element exists
+template<class A>
+inline bool IterableBitset<A>::exists_safe(size_t v) {
```
giovannic, line 397: Exists is already safe. It uses `.at`. Perhaps it's best to catch the C++ error in R and replay your nicer error.

R/bitset.R
```diff
@@ -43,7 +43,15 @@ Bitset <- R6::R6Class(
       self
     },
 
-    #' @description get the number of elements in the set
+    #' @description check if elements are in the Bitset 
+    #' @param v an integer vector of elements to check. If an element of \code{v} is 
+    #' outside the maximum size of the Bitset this is considered an error.
+    #' @return a logical vector the same length as \code{v}
+    exists = function(v) {
```
giovannic, line 50: I'm not sure of the use case for the logical vector.

But it feels neater to have a `to_logical` method.

And then if you really want, you can have:

```r
 exists = function(v) Bitset$new(self$max_size)$insert(v)$and(self)$to_logical()
```

R/bitset.R
```diff
@@ -43,7 +43,15 @@ Bitset <- R6::R6Class(
...
+      bitset_exists_vector(b = self$.bitset, v = as.integer(v))
```
giovannic, line 51: As mentioned below. You could have a tryCatch here.

src/utils.h
giovannic: Oh, it's right above!

giovannic (changes requested):
Hey,

So it would be cool to understand the use case for returning the logical vector.

Also it seems there's a messy git history here. Would it be possible to introduce this as 1 commit from dev? There seem to be 37 commits from a bunch of other familiar branches. It could get confusing.

Other comments inline.

Thanks!

slwu89:
> Hey,
> 
> So it would be cool to understand the use case for returning the logical vector.
> 
> Also it seems there's a messy git history here. Would it be possible to introduce this as 1 commit from dev? There seem to be 37 commits from a bunch of other familiar branches. It could get confusing.
> 
> Other comments inline.
> 
> Thanks!

Yeah it is super messy ugh. Not sure how I got to that point. Actually @giovannic I'll close this PR without merging and clean it up on my end. The overload of `decrement` for Rcpp::IntegerVector input also snuck in.

The use case is that I have another package linking to individual and a network package and needed to be able to check if bits were set. You know now that I think about it, the only actual thing I would need to make that work is for `IterableBitset::exists` to be a public method; what's your opinion on that change? If that's ok, then we can just ignore this PR and I can get on in my other repo.

giovannic:
> the only actual thing I would need to make that work is for IterableBitset::exists to be a public method; what's your opinion on that change?

Of course. lmk. I imagine that could get quite costly if you have to call it several times though.

It'd be easier to review in the context of the network algorithm.

> The overload of decrement for Rcpp::IntegerVector

Oh I seeee

> If that's ok, then we can just ignore this PR and I can get on in my other repo.

Sounds good

slwu89:
> Of course. lmk. I imagine that could get quite costly if you have to call it several times though.
> 
> It'd be easier to review in the context of the network algorithm.

The place I used it is here; basically for each infectious person, we find their contacts, then check which ones are susceptible, add to a vector counting infectious contacts. https://github.com/slwu89/individual/blob/feat/network/src/network.cpp#L50

Of course maybe it would just be better to add all infectious contacts to that vector regardless of the status of the recipient and then subset it with the bitset for susceptible persons. Programming Is Fun.

PR #94, opened 2021-05-17
check for non-finite values in bitset$sample

slwu89:
In the current version of individual, the sample method for bitsets will treat a vector of `NA` as being all 1s. I think it's because `storage.mode(c(NA,NA))` returns "logical" so when it gets to the C++ it is coerced to 1.0 because that vectors type is `double`, giving the observed behavior below. This PR adds a simple check to the sample method for vectors to make sure all the inputs are ok, and an additional test for this. This isn't a problem for a scalar `rate`, so we only need to test for a vector input.

This is relevant when the probabilities vary by timestep (for example), if the timestep is greater than the length of some vector of probabilities, R will return NA which could be passed to bitset$sample. Because no error is thrown this can lead to inscrutable output, so checking can lead to faster finding of bugs.

```
> library(individual)
> bset <- Bitset$new(size = 10)$insert(1:10)
> bset$to_vector()
 [1]  1  2  3  4  5  6  7  8  9 10
> bset$sample(rate = rep(NA,10))
> bset$to_vector()
 [1]  1  2  3  4  5  6  7  8  9 10
```

### DIFF
--- a/R/bitset.R
+++ b/R/bitset.R
@@ -88,6 +88,7 @@ Bitset <- R6::R6Class(
       if (length(rate) == 1) {
         bitset_sample(self$.bitset, rate)
       } else {
+        stopifnot(all(is.finite(rate)))
         bitset_sample_vector(self$.bitset, rate)
       }      
       self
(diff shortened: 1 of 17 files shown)


giovannic (approved):
Thanks!

PR #96, opened 2021-05-20
Updates to Performance vignette

slwu89:
Adds a section to demonstrate the bitset$sample() and bitset$set_difference() pattern being faster than calling filter_bitset() twice. It's a pretty common pattern so worth showing people up front. Also add the step of creating src/Makevars specifying the c++14 standard to the guide on how to make a package that links with "individual"; without that it won't compile.

### DIFF
--- a/vignettes/Performance.Rmd
+++ b/vignettes/Performance.Rmd
@@ -63,9 +63,33 @@ infection_process <- function(t){
 
 When creating a new Bitset, a user must specify the maximum size of the bitset. This is the maximum number of positive integers which the bitset can store. For example, if calling `Bitset$new(size = 100)`, the resulting object is able to store the presence or absence of integers between 1 and 100 (inclusive). Attempting to insert or remove elements outside of this range will result in an error.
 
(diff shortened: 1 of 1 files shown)


vignettes/Performance.Rmd
```diff
@@ -63,9 +63,33 @@ infection_process <- function(t){
...
+leave <- filter_bitset(bitset = bset,other = which(!keep))
+```
+
+This pattern is almost always slower than using the sample method with a set difference:
+
+```{r,eval=FALSE}
+stay <- bset$copy()
+stay$sample(rate = probs)
+
+leave <- bset$copy()$set_difference(stay)
```
giovannic, line 87: This is so cool. I'm gonna use this in a talk.

[not essential] It would be cool to have some microbenchmarks on this page too. It would make it even more compelling, e.g.

```r
microbenchmark(
  mean1(x),
  mean2(x)
)
#> Unit: nanoseconds
#>      expr   min    lq   mean median    uq        max neval
#>  mean1(x) 4,670 4,860  26557  5,020 5,220  2,150,000   100
#>  mean2(x)   968 1,120 204006  1,340 1,420 20,300,000   100
```

giovannic (approved):

vignettes/Performance.Rmd
slwu89: I agree. I wasn't sure if we wanted to have microbenchmarks running in the vignette but I do think it would help make the case for using those class methods (I did my own benchmarks to convince myself what I was writing is true...after about 1000 individuals the sample()/set_difference() pattern becomes 2-4X faster on my system).

PR #98, opened 2021-05-25
consistent use of stl algorithms for IntegerVariable and DoubleVariable

slwu89:
This PR changes the methods of IntegerVariable and DoubleVariable for finding size of sets and getting indices to consistently use functions in the <algorithm> header, for consistency. In a few cases (particularly for IntegerVariable) there is a small performance benefit but mostly this is to enforce consistency in the implementations. Also adds missing test cases for IntegerVariable in the test suite.

Somehow the dev branch had a misspelled roxygen tag in render.R so to avoid checks failing this PR fixes that too.

### DIFF
--- a/inst/include/DoubleVariable.h
+++ b/inst/include/DoubleVariable.h
@@ -56,11 +56,13 @@ struct DoubleVariable : public Variable {
     ) const {
         
         auto result = individual_index_t(size);
-        for(size_t it = 0; it < values.size(); it++) {
-            if( !(values[it] < a) && !(b < values[it]) ) {
-                result.insert(it);
+        auto i = 0u;
+        std::for_each(values.begin(), values.end(), [&](const double v) -> void {
+            if( !(v < a) && !(b < v) ) {
+                result.insert(i);
             }
-        }
+            ++i;
+        });
         
         return result;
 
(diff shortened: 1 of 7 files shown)


inst/include/DoubleVariable.h
```diff
@@ -71,12 +73,9 @@ struct DoubleVariable : public Variable {
         const double a, const double b
     ) const {
         
-        int result{0};
-        for(size_t it = 0; it < values.size(); it++) {
-            if( !(values[it] < a) && !(b < values[it]) ) {
-                result += 1;
-            }
-        }
+        int result = std::count_if(values.begin(), values.end(), [&](const double v) -> bool {
```
giovannic, line 76: Can we use an unsigned integer for counts?

```suggestion
        size_t result = std::count_if(values.begin(), values.end(), [&](const double v) -> bool {
```

it's, again, more explicit about what we expect from that line

slwu89:
Thanks for the feedback @giovannic! I did some more serious benchmarking in the meantime and found that the `for_each` pattern doesn't make any difference in speed (for some reason for DoubleVariable it slows things down, but not for IntegerVariable? mysterious ways).  `count_if` on the other hand is a bit faster. 

How about we ditch the `for_each` implementations, as those do indeed get pretty cumbersome, but keep the `count_if` ones, which I think are pretty convenient to write anyway.

slwu89:
Made those changes @giovannic! Also the IntegerVariable method to count the number of individuals with some specific value wasn't dispatching to the cpp implementations based on whether it gets scalar or vector input. So the `IntegerVariable$get_size_of` method now chooses the right implementation. Even with the extra check for length of input at the R level, avoiding using `std::find` within the loop speeds things up a lot for reasonable population sizes (1e5). 

image

giovannic (approved):

giovannic:
Great work!

Issue #104, opened 2021-06-15
randomly sample bitset given fixed number of elements to retain


slwu89:
@giovannic for something in safir I realized it would be nice to add a method for `Bitset` objects that would let you randomly sample a `N` number of indices to retain. It can be done somewhat awkwardly right now like this:
```
n <- 100
bset <- Bitset$new(n)$insert(1:n)
num_to_keep <- 30
to_remove <- sample.int(n = n,size = n - num_to_keep,replace = FALSE)
bset$remove(to_remove)
bset$size() == num_to_keep
```
What do you think about making some function that acts on the `IterableBitset` where you tell it `N`, the number of items to retain in the bitset, and then it will randomly sample the indices of the ones to keep? (and accompanying R interface)? I imagine it could be virtually identical to `bitset_sample_internal` except the 2nd argument to `Rcpp::sample` would be `b.size() - N`.
https://github.com/mrc-ide/individual/blob/18ddd6a84a8fc94d7920d71f5cb523764e971df3/inst/include/IterableBitset.h#L436-L464

An important question is what to call this method..."random_subset", "keep", or something else? Also, I just realized this could be accomplished with `filter_bitset`, but the performance benefit of a modifying method might be worth it.

slwu89:
Cool. I'll work something out, especially since it would be useful for safir. For the R6 method, is there a name for this operation you'd like? sample is taken..., subset, keep, random_subset? Open to suggestions (because I don't know what method name clearly communicates to people what this operation is doing).

giovannic:
Ah yes. random_subset seems like the most descriptive name there.

PR #111, opened 2021-06-16
random_subset method for bitsets

slwu89:
This PR should address #104 with a new method `random_subset`; there is not a ton of checking done at the R level because `Rcpp::sample` gives pretty informative error messages for bad inputs.

Still happy to change the method name if we come up with something better. `choose` is one option, since it brings to mind the binomial "n choose k".

### DIFF
--- a/tests/testthat/test-bitset.R
+++ b/tests/testthat/test-bitset.R
@@ -173,3 +173,31 @@ test_that("bitset throws error when given bad input probabilities in sample", {
     b$sample(rate = c(rep(0.1,9),NA))
   )
 })
+
+test_that("bitset random_subset behaves properly when given an empty bitset", {
+  b <- Bitset$new(10)
+  expect_error(
+    b$random_subset(5)
+  )
+  expect_error(
+    b$random_subset(-1)
+  )
+  expect_error(
+    b$random_subset(100)
+  )
+  expect_error(
+    b$random_subset(Inf)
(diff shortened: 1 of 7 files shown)


tests/testthat/test-bitset.R
```diff
@@ -173,3 +173,31 @@ test_that("bitset throws error when given bad input probabilities in sample", {
...
+  )
+  expect_error(
+    b$random_subset(Inf)
+  )
+})
+
+test_that("bitset random_subset behaves properly when given a bitset with elements", {
+  
+  b <- Bitset$new(10)$insert(1:8)
+  expect_equal(b$random_subset(5)$size(), 5)
```
giovannic, line 196:
```suggestion
  expect_equal(b$random_subset(5)$size(), 5)
  expect_equal(b$copy()$or(b$copy()$random_subset(5))$to_vector(), b$to_vector()) # check that b$random_subset is a subset of b
```

giovannic (changes requested):
I quite like `choose` actually! It's up to you :)

tests/testthat/test-bitset.R
slwu89: added this test! thanks, that was a good catch. Also renamed `random_subset` -> `choose`, I think it's the most descriptive so far.

giovannic (approved):

PR #112, opened 2021-06-21
Paper/sof

giovannic:
* Reduced State of Field in paper to focus on R based Epi, IBM packages like our own. (removed 381 words)
 * Added alternatives to the README
 * Fixed a few typos


slwu89 (approved):
Thanks @giovannic, that looks great. After this is merged I can get rid of the example section and instead add links to the SIR tutorial in the vignettes.

Issue #118, opened 2021-07-13
CRAN's LTO check issue


slwu89:
@giovannic I don't know if this is contributing to the issues coming up in the CRAN checks but I am seeing this coming up from building on Rhub. Was on this platform: Windows Server 2008 R2 SP1, R-release, 32/64 bit

```
In file included from ../inst/include/common_types.h:15,
                 from bitset.cpp:10:
../inst/include/IterableBitset.h: In instantiation of 'void bitset_sample_internal(IterableBitset<A>&, double) [with A = long long unsigned int]':
bitset.cpp:99:42:   required from here
../inst/include/IterableBitset.h:485:22: warning: comparison of integer expressions of different signedness: 'unsigned int' and 'int' [-Wsign-compare]
       while(bitset_i != i) {
             ~~~~~~~~~^~~~
../inst/include/IterableBitset.h: In instantiation of 'void bitset_choose_internal(IterableBitset<A>&, size_t) [with A = long long unsigned int; size_t = unsigned int]':
bitset.cpp:164:33:   required from here
../inst/include/IterableBitset.h:455:20: warning: comparison of integer expressions of different signedness: 'unsigned int' and 'int' [-Wsign-compare]
     while(bitset_i != i) {
           ~~~~~~~~~^~~~
```
It's coming from `bitset_sample_internal` and `bitset_choose_internal`. My guess is that the object returned from `Rcpp::sample` [in this line](https://github.com/mrc-ide/individual/blob/master/inst/include/IterableBitset.h#L474) isn't the right type to be compared with `bitset_i`, and because the iterator variable `i` "inherits" (sorry for loose language, I have no idea how `auto` type deduction works) from the return type of `Rcpp::sample`, we see the comparison between signed and unsigned types. That's my hypothesis at least.

The full build log is here: [00install.txt](https://github.com/mrc-ide/individual/files/6810603/00install.txt)

slwu89:
Another warning coming up on platform: macOS 10.13.6 High Sierra, R-release, CRAN's setup

```
clang++ -mmacosx-version-min=10.13 -std=gnu++14 -I"/Library/Frameworks/R.framework/Resources/include" -DNDEBUG  -I'/Users/user79uAzzgl/R/Rcpp/include' -I'/Users/user79uAzzgl/R/testthat/include' -I/usr/local/include   -fPIC  -Wall -g -O2  -c bitset.cpp -o bitset.o
In file included from bitset.cpp:10:
[...]

PR #121, opened 2021-07-16
CRAN test fixes:

giovannic:
* fix signed compiler warnings on certain platforms
 * fix suspected UBSAN issue (cannot recreate on rocker)
 * formatting fixes for tests

### DIFF
--- a/R/double_variable.R
+++ b/R/double_variable.R
@@ -85,6 +85,7 @@ DoubleVariable <- R6Class(
           index <- index$to_vector()
         }
         if (length(index) != 0) {
+          stopifnot(all(index > 0))
           double_variable_queue_update(
             self$.variable,
             values,
--- a/R/integer_variable.R
+++ b/R/integer_variable.R
@@ -117,6 +117,7 @@ IntegerVariable <- R6Class(
           index <- index$to_vector()
         }
         if (length(index) != 0) {
+          stopifnot(all(index > 0))
           integer_variable_queue_update(
             self$.variable,
             values,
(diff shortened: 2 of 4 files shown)


slwu89:
looks good! this will close out #118.

R/double_variable.R
```diff
@@ -85,6 +85,7 @@ DoubleVariable <- R6Class(
           index <- index$to_vector()
         }
         if (length(index) != 0) {
+          stopifnot(all(index > 0))
```
slwu89, line 88: Is this necessary? Currently without this addition when I queue an update with negative indices it gives a reasonable error:
```
> x <- IntegerVariable$new(initial_values = 1:10)
> x$queue_update(values = 1,index = -5:-1)
Error in integer_variable_queue_update(self$.variable, values, index) : 
  Index out of bounds
```

R/integer_variable.R
```diff
@@ -117,6 +117,7 @@ IntegerVariable <- R6Class(
           index <- index$to_vector()
         }
         if (length(index) != 0) {
+          stopifnot(all(index > 0))
```
slwu89, line 120: same Q as for double_variable.R

slwu89 (commented):
Nice, we can close #118 with this. Had a question about the changes to IntegerVariable and DoubleVariable though.

R/double_variable.R
slwu89: Oh that makes sense, I hadn't seen that error log. Nice catch!

PR #124, opened 2021-07-20
IntegerVariable get methods now allow a=b which defaults to set behavior

slwu89:
Sometimes it is useful to call `IntegerVariable$get_size_of(a = val1, b = val2)` in a modeling context where it is possible for `val1 = val2`; right now this is considered an error and the code requires `a < b`. I propose that in this case (`a = b`) the method should revert to the set behavior with a set consisting of a single element `integer_variable_get_index_of_set_scalar`. This PR implements that for both `get_size_of` and `get_index_of` methods and adds another test.


giovannic (approved):
Looks great

PR #126, opened 2021-07-22
Hotfix/cran checks

slwu89:
fixed one last unsigned/signed comparison that was giving LTO error in IterableBitset.h `bitset_choose_internal`. No further errors/warnings/notes after running [rhub checks](https://r-hub.github.io/rhub/reference/check_shortcuts.html) `check_with_sanitizers` and `check_with_valgrind`  both with option `--enable-lto=check` enabled (as per the [R manual on writing extensions](https://cran.r-project.org/doc/manuals/r-devel/R-admin.html#LTO-for-package-checking))

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: individual
 Title: Framework for Specifying and Simulating Individual Based Models
-Version: 0.1.3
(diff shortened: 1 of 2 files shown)


giovannic (approved):
Incredible!

PR #134, opened 2021-08-09
improvements to variable classes, clean up code and tests

slwu89:
@giovannic this PR improves the variable classes, and addresses some other issues I've found over the past few weeks. I've checked this branch using R-CMD-check as well as with the docker containers for R compiled with extra options to catch undef behavior and valgrind and everything passes without any problems (no undefined behavior or memory leaks detected).

Sorry for the huge PR...I let things get out of hand. I'll list below a summary of major changes and we'll see what you think and how many changes should stay:

1. remove `reshape2` from DESCRIPTION: we don't seem to use this package anywhere in `individual`, and all checks pass after removal.
2. in header files for `CategoricalVariable`, `IntegerVariable`, and `DoubleVariable`, move all method definitions to outside the struct (the same style as done for `IterableBitset`; I find this helps me see more clearly what methods the struct supports (and their argument overloads), and then I can look at specific implementation below if needed ([example for categorical variable](https://github.com/mrc-ide/individual/blob/7692c111ae21b87117858a986b67d43ed6247b74/inst/include/CategoricalVariable.h)).
3. added pseudo-docs tags to the methods in `CategoricalVariable`, `IntegerVariable`, and `DoubleVariable` headers, to match the style of `IterableBitset`
4. `get_values` and `get_size` methods for `IntegerVariable`, and `DoubleVariable` were returning a reference, fixed here to return an object https://github.com/mrc-ide/individual/commit/9b48a3608e68f50452e78bc2ff341e37840309b2
5. give `CategoricalVariable`, `IntegerVariable`, and `DoubleVariable` a virtual destructor with default behavior https://github.com/mrc-ide/individual/commit/ea8f9b0d3382ffc8e25b21b437233658563d8503
6. `get_size_of` methods for `CategoricalVariable` were returning `int`, change to return `size_t` for consistency with the other variable types.
7. simplified `get_index_of` method for `CategoricalVariable` when the user passes a single category, no need to make a empty bitset and do bitset or, just use bitset constructor with the existing bitset attached to that category as argument.
8. remove `#include <bitset>` from `IterableBitset.h` and `#include <sstream>` from other variable headers.
9. several for loops check if a type `size_t` is less than 0; removed these unnecessary comparisons (https://github.com/mrc-ide/individual/commit/8205290d3f503bdb91b2305e672b95dea4361b39, and https://github.com/mrc-ide/individual/commit/995186c3b970ff050040286cdb1654e2d8460ac1)
10. if passed an integer vector for index, `get_values` method for `IntegerVariable` can directly retrieve values with the vector rather than make a corresponding bitset and then get values using the bitset. This is moderately faster. [intvar_get_values_at_index_vector.pdf](https://github.com/mrc-ide/individual/files/6956236/intvar_get_values_at_index_vector.pdf)
11. if passed an integer vector for `index`, `queue_update` for `IntegerVariable`, and `DoubleVariable` now does the conversion from vector to bitset in C++. It's very slightly faster [intvar_queue_update_bitset.pdf](https://github.com/mrc-ide/individual/files/6956242/intvar_queue_update_bitset.pdf)
12. It's been useful to convert bitsets to vectors for some cases in safir; I added a function to `IterableBitset` that just makes it more convenient, rather than typing the loop each time https://github.com/mrc-ide/individual/blob/3de72096b9e940056df35efd278f7df751976050/inst/include/IterableBitset.h#L408
13. Split the test file `test-variables.R` into 3 files, one for each variable type. Add more tests for edge cases that were not being caught. The single file was getting confusing to look at IMO.
14. Pass arguments by reference where reasonable in `IntegerVariable` and `DoubleVariable`: https://github.com/mrc-ide/individual/commit/65c0a7819a76ecba8db72dee068113faa196b172


giovannic (approved):
Looks great!

PR #145, opened 2021-09-14
check for bad bitset index when queuing CategoricalVariable updates 

slwu89:
Currently the following code does not cause an error, in contrast to how IntegerVariable and DoubleVariable handle bad bitset input when queuing updates:  
```
c <- CategoricalVariable$new(categories = c("A","B"),initial_values = rep(c("A","B"),each=10))
c$queue_update(value = "A",index = Bitset$new(50)$insert(c(15, 25, 50)))
```

It's nicer to make the error handling consistent, as then the error can be traced more easily back to the process that queued the bad indices, rather than starting from when it tried to apply the update. This PR also adds some tests to catch these cases.


giovannic (approved):
Great. Thanks!

PR #159, opened 2021-11-12
Feat/resizable

giovannic:
Some simulations require dynamic population sizes, this is a start at implementing that support. Hopefully the first of three pull requests for resizeable variables (Double, Integer and Categorical). This one is just for ResizeableDoubleVariable.

ResizeableDoubleVariable allows clients to queue "extentions" and "shrinks", which will be applied after normal updates. 

Extentions will add new individuals with an initial value. e.g.

```r
age$queue_extend(rep(0, n_births))
```

It can be used to model births or importation in a population.

Shrinks can selectively remove individuals from a variable, using either a bitset or a variable. e.g.

```r
age$queue_shrink(deaths)
```

It can be used to model death-like processes in a population.

Notable Changes:

 * Consolidated filtering algorithm (from bitset_filter) into FilterIterator, so that it can be used with any iterables
 * New C++ classes, C++ interface and R6 objects for ResizeableDoubleVariable
 * Inheritance-based resizing updates


R/resizeable_double.R
```diff
@@ -0,0 +1,147 @@
+#' @title DoubleVariable Class
```
slwu89, line 1: Can you update the `DoubleVariable` to `ResizeableDoubleVariable` in the class docs?

R/resizeable_double.R
```diff
@@ -0,0 +1,147 @@
...
+#' @description Represents a continuous variable for an individual.
```
slwu89, line 2: It would be nice to give the update order, i.e. when are resizing updates applied, so users can structure processes correctly.

src/utils.h
```diff
@@ -5,15 +5,13 @@
  *      Author: gc1610
  */
 
-#ifndef UTILS_H_
```
slwu89, line 8: perhaps we can move this code to inst/include/utils.h?

slwu89 (changes requested):
@giovannic this looks good, I just have a few comments.

R/resizeable_double.R
giovannic: Done. I'll add a vignette at the end of this epic too

slwu89: oh no not jira

R/resizeable_double.R
```diff
@@ -0,0 +1,149 @@
...
+        } else {
+          resizeable_double_variable_queue_update(
+            self$.variable,
+            values,
+            integer(0)
+          )
+        }
+      } else {
+        if (inherits(index, 'Bitset')) {
+          if (index$size() > 0){
```
slwu89, line 91: Sorry, I didn't catch this on the last review. I think we want to check here to make sure the bitset max size is the same as the variable size, the changes I made in #156. You might do it in the cpp but it would be nice to have it standardized across variable types. I check at the R level for "variable reset" and "subset update with bitset" type updates: https://github.com/mrc-ide/individual/blob/dev/R/double_variable.R#L74

src/utils.h
slwu89: sounds good.

slwu89 (commented):
Hi @giovannic, sorry it took me so long to get to this. I looked everything over, it looks great, I like the pimpl-style thing you have going on with the R6 variable classes.

PR #160, opened 2021-11-14
fix undef behavior bug

slwu89:
@giovannic I will trade you a PR for a PR =)

Fixing a minor bug that causes undef behavior when testing.


giovannic (approved):
:o good catch

PR #161, opened 2021-12-10
Feat/resize int

giovannic:
To be reviewed after #159 

ResizeableIntegerVariable

 * Consolidate common C++ code
 * Consolidate common R6 code


slwu89 (commented):
@giovannic this is nice, and clearly a monumental amount of work. I have some comments which I've left in the specific files. I also have 2 general comments:

1. Are there any functions that are *required* to go into the VariableInterface$interface, that is do we need to enforce the existence of some interface methods? It might be nice to document it somewhere, or even do checking in the `initialize` method of that class.
2. The design is now sufficiently complex we ought to document it somewhere, as it took me a while to figure out how the interfaces work. I made a rough UML chart here, maybe when this is merged we can figure out a vignette to describe how the design works. [individual_UML.drawio.pdf](https://github.com/mrc-ide/individual/files/7786641/individual_UML.drawio.pdf)

giovannic:
Obsolete

PR #163, opened 2021-12-27
fix bug with cpp event listeners

slwu89:
Fix a simple bug that prevents event listeners written in C++ from executing properly.


slwu89:
Hey @giovannic sorry to do this PR without your review, just a simple 1 line fix, needed it for safir.

PR #166, opened 2022-06-30
Epic/resizable simple

giovannic:
A much simpler alternative to the [other](https://github.com/mrc-ide/individual/tree/epic/resizeable) implementation of resizing.

I've done away with the refactoring of R (with all the interfaces and that) and the extra classes on the C++ end (ResizeableVariable and it's derived classes). The drawbacks of this implementation is that, as before, we have a lot of repeated code between Categorical, Double and Integer variables. And on resizing is likely to be expensive for `shrink` operations since we have an `O(size of the vector)` algorithm for vectors instead of `O(number of removals)` algorithm for lists in the more complex implementation. But on the positive side, the R6 classes for category, integer and double variables are more readable this way (inheritance is quite difficult to implement nicely with R6). And there is much less code on the C++ side too.

Changes:
 * CategoricalVariable, DoubleVariable and IntegerVariable have `$queue_shrink` and `$queue_extend` methods now
 * Change C++ structs -> classes for better encapsulation
 * Add a FilterIterator template for easily filtering out iterables
 * Tests


R/event.R
```diff
@@ -36,7 +36,7 @@ Event <- R6Class(
       for (listener in self$.listeners) {
         if (event_should_trigger(self$.event)) {
           if (inherits(listener, "externalptr")) {
-            self$.process_listener_cpp(listener)
+            self$.process_listener_cpp(self$.event, listener)
```
slwu89, line 39: Should this function take 1 argument or 2? The definition several lines below only has 1 arg.

slwu89 (commented):
Hi @giovannic, I submitted a PR at #167 into this branch for some things I needed to do to get tests to pass on it, and a few missing words here and there. Some of the changes in that PR are actually just to get this branch looking like the current master branch.

inst/include/NumericVariable.h
```diff
@@ -0,0 +1,271 @@
...
+#ifndef INST_INCLUDE_NUMERIC_VARIABLE_H_
+#define INST_INCLUDE_NUMERIC_VARIABLE_H_
+
+#include "Variable.h"
+#include "common_types.h"
+#include <Rcpp.h>
+#include <queue>
+
+template <class A>
+struct NumericVariable;
```
slwu89, line 17: `class` instead of `struct`?

inst/include/CategoricalVariable.h
```diff
@@ -15,41 +15,51 @@
 
 struct CategoricalVariable;
```
slwu89, line 16: `class` instead of `struct`?

inst/include/Variable.h
```diff
@@ -7,8 +7,12 @@
 #ifndef INST_INCLUDE_VARIABLE_H_
 #define INST_INCLUDE_VARIABLE_H_
 
+#include <cstddef>
+
 struct Variable {
```
slwu89, line 12: `class` instead of `struct`?

inst/include/Event.h
```diff
@@ -110,12 +111,15 @@ inline void Event::clear_schedule() {
 //'     * t: current simulation time step
 //'     * targeted_schedule: a map of times and bitsets of scheduled individuals
 //'     * size: size of population
-struct TargetedEvent : public EventBase {
+class TargetedEvent : public EventBase {
```
slwu89, line 114: should `EventBase` and `Event` be `class` definitions instead of `struct` for consistency?

slwu89 (commented):
look good! I just made a few comments regarding `struct` vs `class` consistency in the header files.

giovannic:
Ok, so I made the forward declarations classes for the sake of readability, and EventBase now controls `t`.

But it seems fine to have `Variable` as a struct since it doesn't (and will probably never) have any private members.

slwu89 (commented):
looks good! all the builds work fine on my local system.

PR #171, opened 2022-07-21
Bug/cran memory checks

giovannic:
Fixes the CRAN memory check [here](https://www.stats.ox.ac.uk/pub/bdr/memtests/valgrind/individual/tests/testthat.Rout) and cleans up some duplication.


slwu89 (commented):
Look good to me! R CMD check passes locally.

PR #173, opened 2022-08-26
Bugfix/html errors

slwu89:
I think the errors coming up on CRAN were related to R switching to HTML5 (https://blog.r-project.org/2022/04/08/enhancements-to-html-documentation/) for validation in the newest version (see this issue https://github.com/r-lib/roxygen2/issues/1290). I updated the Roxygen2 version needed and set roxygen2 to use markdown syntax for parsing Rd files. I also simplified the offending parts of TargetedEvent to use markdown style links to help clear things up. Everything now passes on r-hub builders (using https://r-hub.github.io/rhub/reference/check_for_cran.html)


giovannic (approved):
Ahhh thank you!!

Issue #165, opened 2022-05-31
Remove deprecation warning for Bitset$not


giovannic:
Everyone should know by the next version

slwu89:
Shall we go ahead and remove the warning and advance? It seems its been long enough for the knowledge to percolate around.

PR #174, opened 2023-08-29
Bitset not is inplace by default:

giovannic:
Bitset not is inplace by default:

 * Remove deprecation warning
 * Set inplace to TRUE if missing

Solves #165

### DIFF
--- a/R/bitset.R
+++ b/R/bitset.R
@@ -68,17 +68,10 @@ Bitset <- R6Class(
     },
 
     #' @description to "bitwise not" or complement a bitset.
-    #' @param inplace whether to overwrite the current bitset.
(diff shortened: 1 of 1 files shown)


R/bitset.R
```diff
@@ -68,17 +68,10 @@ Bitset <- R6Class(
     },
 
     #' @description to "bitwise not" or complement a bitset.
-    #' @param inplace whether to overwrite the current bitset.
+    #' @param inplace whether to overwrite the current bitset, default = TRUE
     not = function(inplace) {
```
slwu89, line 72: should `inplace = TRUE` in the argument list of the function to implement the default argument behavior?

slwu89 (commented):
Hooray!

Issue #154, opened 2021-10-21
shadow variable benchmark


slwu89:
@giovannic I have an implementation of the shadow integer variable [here](https://github.com/slwu89/individual/blob/feat/variable-enhancements/inst/include/Shadow_IntegerVariable.h). The way it works is pretty straightforward, there are just 2 vectors it contains, each time step an enum is used to track which one is "active" and which is the "shadow". Queuing updates overwrites the shadow. The [update](https://github.com/slwu89/individual/blob/feat/variable-enhancements/inst/include/Shadow_IntegerVariable.h#L275) sets the shadow to be the new active vector by swapping the enum and copies the old shadow to the new shadow vector.

The result of one of the benchmarks is here (the pattern is similar for the rest). As you can see, for large variables when only a few elements are being updated, the current implementation is faster. For larger variables when a high proportion of elements are being updated however, the shadow variable is faster, see limit (variable size) of 1e6  with updating 9e5 elements in the lower right, where the shadow variable is quite a bit faster.

[vv_bi.pdf](https://github.com/mrc-ide/individual/files/7392853/vv_bi.pdf)

So it's a tradeoff = ) Of course the double variable's shadow implementation should look almost identical. It may be less of a tradeoff for the double variable, which is probably updated much more frequently, with more of the population being updated.

PR #176, opened 2023-12-21
Speedup render by avoiding repeated vector copies.

plietar:
The `Render$render` method is used throughout a simulation to store data that needs outputting at the end. It does so by creating a vector per metric, large enough to fit all timesteps.

Unfortunately, each time the `render` method is called, the assignment into the vector actually creates a new copy, instead of modifying it in-place. For long simulations (ie. tens of thousands of timesteps), these vectors are both fairly large and get copied over and over again at each timestep, to the point where the `render` method dominates the execution time.

The core of the problem can easily be reproduced by recreating the same conditions of the `Render` class, using a vector stored inside of an environment:

```r
e <- new.env(parent=emptyenv())
e$vector = rep(0, 10)
tracemem(e$vector)
# [1] "<0x56313d8a4c80>"
e$vector[[0]] <- 1
# tracemem[0x56313d8a4c80 -> 0x56314180ff60]:
e$vector[[1]] <- 1
# tracemem[0x56314180ff60 -> 0x563141613fc0]:
e$vector[[2]] <- 1
# tracemem[0x563141613fc0 -> 0x563141984b10]:
```

Here `e` mirrors the `private` environment of the `Render` class, as it would be constructed by R6. The `tracemem` call is used to trace copies of the vector. Each subsequent assignment to the vector creates a new copy of it, which can be seen as tracemem lines in the console.

I did experiment with ways of avoiding the copies without resorting to C++ code, but I couldn't find anything that wasn't made of horrible hacks.

The patch here fixes this issue by creating a C++ class, that acts as a thin wrapper around a vector. Thanks to this indirection, the vector is guarateed to have reference semantics and be modified in-place. A method exists on the wrapper class to extract the underlying vector. This does incur a copy, but is only expected to be called once (per vector), at the end of the simulation.

For large runs, this change has a dramatic effect of the performance of the end-to-end simulation. Using malariasimulation as a benchmark:
```r
# Before the change
system.time(run_simulation(timesteps = 50000))
#    user  system elapsed
# 334.662   0.604 335.253

# Afterwards
system.time(run_simulation(timesteps = 50000))
#    user  system elapsed
# 134.270   0.008 134.271
```

The impact is much less noticeable are shorter runs, since the vectors being copied are much smaller (on the order of 20% improvement for a 10'000 time steps run, 5% for 5'000 time steps).

### DIFF
--- a/R/RcppExports.R
+++ b/R/RcppExports.R
@@ -437,6 +437,18 @@ integer_ragged_variable_queue_shrink_bitset <- function(variable, index) {
     invisible(.Call(`_individual_integer_ragged_variable_queue_shrink_bitset`, variable, index))
 }
 
+create_render_vector <- function(data) {
+    .Call(`_individual_create_render_vector`, data)
+}
+
+render_vector_update <- function(v, index, value) {
+    invisible(.Call(`_individual_render_vector_update`, v, index, value))
+}
+
+render_vector_data <- function(v) {
+    .Call(`_individual_render_vector_data`, v)
+}
+
 execute_process <- function(process, timestep) {
     invisible(.Call(`_individual_execute_process`, process, timestep))
 }
@@ -455,5 +467,5 @@ variable_resize <- function(variable) {
 
 # Register entry points for exported C++ functions
 methods::setLoadAction(function(ns) {
-    .Call('_individual_RcppExport_registerCCallable', PACKAGE = 'individual')
+    .Call(`_individual_RcppExport_registerCCallable`)
 })
--- a/R/render.R
+++ b/R/render.R
@@ -16,7 +16,7 @@ Render <- R6Class(
     #' @param timesteps number of timesteps in the simulation.
     initialize = function(timesteps) {
       private$.timesteps = timesteps
-      private$.vectors[['timestep']] <- seq_len(timesteps)
+      private$.vectors[['timestep']] <- create_render_vector(seq_len(timesteps))
     },
     
     #' @description
@@ -28,7 +28,7 @@ Render <- R6Class(
       if (name == 'timestep') {
         stop("Cannot set default value for variable 'timestep'")
       }
-      private$.vectors[[name]] = rep(value, private$.timesteps)
+      private$.vectors[[name]] = create_render_vector(rep(value, private$.timesteps))
     },
 
     #' @description
@@ -41,15 +41,15 @@ Render <- R6Class(
         stop("Please don't name your variable 'timestep'")
       }
       if (!(name %in% names(private$.vectors))) {
-        private$.vectors[[name]] = rep(NA, private$.timesteps)
+        private$.vectors[[name]] = create_render_vector(rep(NA, private$.timesteps))
       }
-      private$.vectors[[name]][[timestep]] = value
+      render_vector_update(private$.vectors[[name]], timestep, value)
     },
 
     #' @description
     #' Return the render as a \code{\link[base]{data.frame}}.
     to_dataframe = function() {
-      data.frame(private$.vectors)
+      data.frame(lapply(private$.vectors, render_vector_data))
     }
   )
 )
--- a/R/render.R
+++ b/R/render.R
@@ -15,7 +15,7 @@ Render <- R6Class(
     #' renderers.
     #' @param timesteps number of timesteps in the simulation.
     initialize = function(timesteps) {
-      private$.timesteps = timesteps
+      private$.timesteps <- timesteps
       private$.vectors[['timestep']] <- create_render_vector(seq_len(timesteps))
     },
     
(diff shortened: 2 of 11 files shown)


src/render_vector.cpp
```diff
@@ -0,0 +1,26 @@
...
+ */
+
+
+#include "../inst/include/RenderVector.h"
+#include <Rcpp.h>
+
+
+//[[Rcpp::export]]
+Rcpp::XPtr<RenderVector> create_render_vector(Rcpp::NumericVector data) {
+    return Rcpp::XPtr<RenderVector>(new RenderVector(data), true);
```
richfitz, line 15: I am not 100% sure about the semantics here of saving a `Rcpp::NumericVector` within an `XPtr`; this might work but might also be something that bites us later. My inclination would have been to use `std::vector<double>` for the storage behind the pointer and dump into `Rcpp::NumericVector` on export (so definitely one copy but also no risk of later in-place mutation)

richfitz (changes requested):

src/render_vector.cpp
```diff
@@ -0,0 +1,26 @@
...
+}
+
+//[[Rcpp::export]]
+void render_vector_update(Rcpp::XPtr<RenderVector> v, size_t index, float value) {
```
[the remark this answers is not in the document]
plietar: Yep. Thanks.

R/render.R
```diff
@@ -28,7 +28,7 @@ Render <- R6Class(
       if (name == 'timestep') {
         stop("Cannot set default value for variable 'timestep'")
       }
-      private$.vectors[[name]] = rep(value, private$.timesteps)
+      private$.vectors[[name]] = create_render_vector(rep(value, private$.timesteps))
```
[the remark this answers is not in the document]
plietar: Sure. Done.

R/render.R
```diff
@@ -41,15 +41,15 @@ Render <- R6Class(
         stop("Please don't name your variable 'timestep'")
       }
       if (!(name %in% names(private$.vectors))) {
-        private$.vectors[[name]] = rep(NA, private$.timesteps)
+        private$.vectors[[name]] = create_render_vector(rep(NA, private$.timesteps))
```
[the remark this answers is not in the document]
plietar: Applied the suggestion.

At one point I had it with the constructor taking a length, but then I also needed a different constructor for `set_default`. I could have something with a value and length, but just passing a vector and letting the caller use `rep` was simplest.

src/render_vector.cpp
plietar: `std::vector<double>` had the downside of not supporting `NA`, so I thought we would either need to change the semantics of the class, or do some magic to represent the `NA` somehow, whereas by using a `NumericVector` would avoid having to deal with this.

Looking through the R internals docs just now, it turns out that internally R actually just represents `NA` simply as a NaN with a specific bit pattern, so it might just work even stored in a C++ `double`. I'll give it a try and see if that works

src/RcppExports.cpp
```diff
@@ -366,7 +366,7 @@ RcppExport SEXP _individual_dummy() {
     if (rcpp_isError_gen) {
         SEXP rcpp_msgSEXP_gen = Rf_asChar(rcpp_result_gen);
         UNPROTECT(1);
-        Rf_error(CHAR(rcpp_msgSEXP_gen));
+        Rf_error("%s", CHAR(rcpp_msgSEXP_gen));
```
plietar, line 369: This was introduced by running against a prerelease of Rcpp to avoid a CI failure, as explained in https://github.com/RcppCore/Rcpp/issues/1287#issuecomment-1829886024.

We only need the pre-release to generate this file. Users of the package don't need it.

src/render_vector.cpp
plietar: Alright the `std::vector<double>` worked without any special effort needed, including handling NA values (this is covered by the unit tests). I've switched to that.

R/RcppExports.R
```diff
@@ -455,5 +467,5 @@ variable_resize <- function(variable) {
 
 # Register entry points for exported C++ functions
 methods::setLoadAction(function(ns) {
-    .Call('_individual_RcppExport_registerCCallable', PACKAGE = 'individual')
+    .Call(`_individual_RcppExport_registerCCallable`)
```
plietar, line 470: This change comes from an Rcpp change at https://github.com/RcppCore/Rcpp/pull/1256

giovannic (approved):
This is great thanks!

I would suggest you to have a look at (and extend if you want?) the existing [microbenchmarks](https://github.com/mrc-ide/individual/tree/master/tests/performance). They run a variety of scenarios to show the effects of an improvement in a variety of scenarios, [here's](https://github.com/mrc-ide/individual/issues/154#issuecomment-949018319) an example of how it can be helpful.

plietar:
Thanks. I've added two microbenchmarks, one which does a single `render()` call and one which calls it as many times as there are timesteps, mimicking the behaviour of a full simulation.

[Rplots.pdf](https://github.com/mrc-ide/individual/files/13754394/Rplots.pdf)

richfitz (approved):

PR #178, opened 2024-01-08
Add initial checkpoint and restore functionality.

plietar:
This allows a user to run a simulation for a number of time steps and save the state of the simulation. The state can then be restored to continue the simulation.

This is,for example, useful to compare the impact of an intervention based on different parameters. The simulation can be run up to the point of an intervention, when the parameters' value don't matter. The simulation can then be resumed multiple times but with different intervention parameters each time.

For now, only variables are supported. Events will be added next.


plietar:
I still need to add some unit tests to make sure the checkpoint/restore works correctly, but I'm interested in getting some feedback in the meantime.

The long term goal is to support https://github.com/mrc-ide/malariasimulation/issues/196. There's obviously still a long way to go for this, especially since malariasimulation maintains quite a lot of state that isn't visible to `individual`, but this is the first step in that direction.

giovannic (changes requested):
Very minor suggestions, I love the way this is done.

I'm curious about whether users would want to `restore_state` themselves, instead of simulation_loop doing it for them. That way users can do their own (re-)initialisations based on the restored variables. e.g. in malariasimulation, they may want to calculate an FOIM based on restored variables so that they can initialise their differential equation solver.

I'm sure those issues could be worked around by saving more state. My guess is that saving more state would be the most concise approach. But it's something to bear in mind.

Thanks so much!

plietar:
Thanks for the review, I've applied the suggestions.

I think it would be much simpler to have malariasimulation save and restore its state buffers than re-creating them from the Variables (I'm not sure that would even be possible). It should be able to do that before/after calling `simulation_loop`, and let `simulation_loop` call `restore_state` to save the events and variables.

Another option would be to add an argument to `simulation_loop` for custom checkpoint-able objects, and have `restore_state` call `checkpoint`/`restore` on these automatically. Then the caller only has to define a pair of methods on each object but not worry about calling them.

giovannic (approved):
To me, the former option seems like a good place to start. Looking forward to it!

PR #179, opened 2024-01-11
Support checkpoint and restore of Event and TargetedEvent.

plietar:


tests/testthat/test-checkpoint.R
```diff
@@ -4,24 +4,36 @@ test_that("deterministic state model is resumable", {
     health <- CategoricalVariable$new(c('S', 'I', 'R'), rep('S', population))
     render <- Render$new(timesteps)
 
-    shift_generator <- function(from, to, rate) {
+    infection <- TargetedEvent$new(population)
```
giovannic, line 7: [nit] It would be nicer to have this as a separate test "deterministic model with events is resumable". So that a failure would point to a more specific issue.

tests/testthat/test-events.R
```diff
@@ -15,23 +15,6 @@ test_that("first event is triggered at t=1", {
   event$.tick()
 })
 
-test_that("first event is triggered at t=1", {
```
giovannic, line 18: Do we not need this anymore?

tests/testthat/test-targetedevent.R
```diff
@@ -526,3 +526,79 @@ test_that("targeted events work for vector delay, bitset target", {
...
+  mockery::expect_called(listener, 1)
+  new_event$.tick()
+
+  #time = 5
+  new_event$.process()
+  mockery::expect_called(listener, 2)
+  expect_targeted_listener(listener, 2, t = 5, target = c(5, 7))
+})
+
+test_that("targeted events are cleared on restored", {
```
giovannic, line 570:
```suggestion
test_that("targeted events are cleared when restored", {
```

giovannic (changes requested):
V. nice

giovannic (approved):

PR #181, opened 2024-01-26
Don't restore the random number generator state by default.

plietar:
It can have suprising side effects by resetting the user global RNG state to a fixed value, and also get in the way of doing stochastic simulations where we actually want to sample many different futures from a given start point.

The only time where a user actually wants to restore the random number generator state is probably when writing tests that compare a continuous run against a resumed one. They can opt-in to that behaviour with a new argument to `simulation_loop` called `restore_random_state`.

### DIFF
--- a/R/simulation.R
+++ b/R/simulation.R
@@ -37,7 +37,8 @@ simulation_loop <- function(
   events = list(),
   processes = list(),
   timesteps,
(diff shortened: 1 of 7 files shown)


giovannic (approved):
Great!

PR #186, opened 2024-02-21
Switch Bitset from an R6 class to a named list.

plietar:
R6 object have significant overhead, especially when instantiating the objects. While this is acceptable for long-lived objects such as events or variables, bitsets are created and destroyed very regularly during simulations.

We can replace our use of an R6 class for `Bitset` with named lists that are intended to look and feel just like the original API, but which significant performance improvement. The reference semantics provided by R6 don't matter in our case, since all mutability happens behind the external pointer.

On malariasimulation, I get a 30-35% performance improvement when using this new implementation of Bitset on population sizes under 10k, and about 10% speedup at 100k.

The object-oriented named list based interface still adds a bit of overhead compared to using the externalptr and Rcpp functions directly, but doing so requires intrusive changes in the use site.


plietar:
Here are some benchmarks of the old version ("R6"), the new one ("list"), and the equivalent code without any wrappers, calling the C++ functions from R directly ("ptr"):
[part1.pdf](https://github.com/mrc-ide/individual/files/14365123/part1.pdf)

https://gist.github.com/plietar/9d07ffb01b9ea7a27a12696b09242435

These compare the performance of doing `Bitset$create(...)`, calling `get_index_of(...)` on a CategoryVariable and finally calling `not()`.

The `create` and `get_index_of` results are a bit all over the place for larger bitsets, and I'm not sure why. The order in which I run the benchmarks makes a difference, the first one generally being the fastest. This is the same data as the previous graphs, but plotting each run over time: [part2.pdf](https://github.com/mrc-ide/individual/files/14365126/part2.pdf) 
Looking at profiles all the time is spent in `memcpy`/`memset`, so I would wager it is just some weird memory performance behaviour.

plietar:
Unfortunately the generated documentation for this is pretty terrible. I need to figure out a good way of keeping the R6 look but with the alternative implementation.

plietar:
Found a workaround for the documentation problem:

Instead of relying on roxygen to do the formatting, we include Rd markup directly in our code.
However we can make it a bit more convenient by using https://roxygen2.r-lib.org/articles/reuse.html#inline-code
We just have one function that outputs the Rd markdown and gets called in the comment block of every method.

It's not the prettiest implementation, but at least the output looks okay.

giovannic (approved):
Very hacky, I respect it 😆 

Since this is localised to Bitsets, this seems a fine practical decision. But if ever reuse this for other R6 objects, we should consider extending [ROxygen](https://cran.r-project.org/web/packages/roxygen2/vignettes/extending.html) to make documentation cleaner.

PR #187, opened 2024-02-27
Optimize some memory copies away.

plietar:
Couple of hotspots found while running malariasimulation with larger population size. This PR gives a 10% speedup when using population size of 1M.

The most suprising change comes from a modification to the signature of Rcpp-exported functions. Some of them used to accept a `const std::vector` as an argument. However when doing so Rcpp creates an extra unexpected copy in the wrapping code. We end up with two copies, once from R memory to an std::vector in the generated code, and a second one when the method is called. If argument is non-const, Rcpp does not create the intermediate object and the copy is avoided.

This behaviour comes from the difference between the `InputParameter` and `ConstInputParameter` classes in Rcpp's [InputParameter.h] file.

Other changes include:
- Return a constant reference from `NumericVariable::get_values`, instead of a copy.
- Use `std::move` inside `NumericVariable::queue_update` instead of a copy.
- Use `reserve` and `push_back` on a vector instead of filling it with zeros to avoid an unnecessary memset.

With this, most operations that work with values require only one copy, where they used to require two or three.

There are a couple more places that could be optimised further by working directly with R vectors instead of `std::vector`, but they are more intrusive changes and don't appear as significantly in malariasimulation profiles anyway. I've left these out for now.

[InputParameter.h]: https://github.com/RcppCore/Rcpp/blob/c63bae6dea4e3994e6f334612c7837fa18ac1e06/inst/include/Rcpp/InputParameter.h

### DIFF
--- a/inst/include/NumericVariable.h
+++ b/inst/include/NumericVariable.h
@@ -39,14 +39,14 @@ public:
     NumericVariable(const std::vector<A>& values);
     virtual ~NumericVariable() = default;
 
-    virtual std::vector<A> get_values() const;
+    virtual const std::vector<A>& get_values() const;
     virtual std::vector<A> get_values(const individual_index_t& index) const;
     virtual std::vector<A> get_values(const std::vector<size_t>& index) const;
 
     virtual individual_index_t get_index_of_range(const A a, const A b) const;
     virtual size_t get_size_of_range(const A a, const A b) const;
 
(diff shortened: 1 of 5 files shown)


giovannic (approved):
Very surprising!

If Rcpp becomes a big bottleneck, it may be worth considering [cpp11](https://github.com/r-lib/cpp11). I haven't seen any execution benchmarks, but suspect the modern implementation may not have the same issues.

PR #188, opened 2024-02-27
Re-generate documentation files.

plietar:
I changed a function signature in mrc-ide/individual#180, but had not regenerated the `.Rd` files.


plietar:
We should maybe have a CI check that the docs are up-to-date to avoid this issue in the future. Surprisingly I can't find any precedent of people doing this I can copy.

giovannic (approved):
Huh, R CMD Check should normally catch these issues with a warning. My guess is ROxygen's poor R6 support...

PR #189, opened 2024-02-28
Optimize the bitset sampling algorithm.

plietar:
The existing bitset sampling implementation works by using a binomial distribution to decide how many bits to keep, randomly chooses the indices of those bits, sorts the vector of indices and finally iterates over all the bits one by one to clear those not contained in the vector.

This can be very inefficient, in particular when sampling over large bit sets with a very low sampling rate. In that case, the list of indices to keep is roughly as large as the bitset itself, and sorting it requires O(nlog(n)) time, which ends up being significant.

Additionally, walking over every single bit, set or not, to be cleared or not, is pretty inefficient as well.

This commit optimizes the implementation through a few methods:
- Instead of sampling and sorting indices to keep, it randomly samples the size of the gaps between two succesful bernouilli trials. This was inspired by the [FastBernoulliTrial] class.
- When the sampling rate is higher than 1/2, it flips the sampling logic and uses the gaps between two unsuccessful trials, minimizing the number of loop iterations.
- Finally, in order to take full advantage of the gap lengths, it is able to quickly scan through the bitset to skip a given number of set bits, calling popcnt once per words rather than looking at each bit, and can clear entire ranges of bits at once by overwriting entire words, rather than masking bits one by one.

This implementation is faster than the previous one for the entire parameter space. The difference is most drastic for very low sampling rates where the new implementation is more than two orders of magnitude faster.

[FastBernoulliTrial]: https://searchfox.org/mozilla-central/rev/a6d25de0c706dbc072407ed5d339aaed1cab43b7/mfbt/FastBernoulliTrial.h

### DIFF
--- a/inst/include/IterableBitset.h
+++ b/inst/include/IterableBitset.h
@@ -86,6 +86,7 @@ public:
     const_iterator end() const;
     const_iterator cend() const;
     void erase(size_t);
+    void erase(size_t start, size_t end);
     const_iterator find(size_t) const;
     template<class InputIterator>
     void insert(InputIterator, InputIterator);
@@ -98,6 +99,7 @@ public:
     bool empty() const;
     void extend(size_t);
     void shrink(const std::vector<size_t>&);
+    size_t next_position(size_t start, size_t n) const;
 };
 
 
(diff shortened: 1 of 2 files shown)


plietar:
I've experimented a lot with different implementations of this, see https://github.com/plietar/individual/blob/all-the-samplings/inst/include/IterableBitset.h.

Here's a benchmark of all the versions, for different sampling rates: size-1M

- `binomial`: use a binomial sample to decide how many elements to remove, pick that number of indices and sort the list. This is the current implementation
- `binomial_opt`: same as `binomial`, except when the rate is less than 1/2 it flips the logic to have less sorting to do.
- `binomial_opt_faster`: same as `binomial_opt`, but it uses a faster implementation to iterate and skip over set bits, and to erase ranges of bits efficiently
- `binomial_opt_faster_bmi2`: same as `binomial_opt_faster`, but it depends on the BMI2 instruction set to skip over bits even faster
- `naive`: draw a new uniform double between [0, 1] for each element of the bitset
- `naive_bulk`: same as `naive`, but the random draws are done as one call to runif
- `skip`: Draw the number of (un)successful bernouilli trials to figure out how many bits to skip, based on [FastBernouilliTrial.h](https://searchfox.org/mozilla-central/rev/a6d25de0c706dbc072407ed5d339aaed1cab43b7/mfbt/FastBernoulliTrial.h)
- `skip_faster`: Same as `skip`, but uses the fast iteration and range erase. This is what this PR implements.
- `skip_faster_bmi2`: Same as `skip_faster`, but uses the BMI2 instruction set

Fast skip and range-erase over bits is the single biggest source of speedup, especially for low/high sampling rates where the gaps are pretty larger. This is beneficial whether the `binomial` or `skip` implementation is used.

The BMI2 instructions gives a non-negligeable speedup, but they are not available by default and require the right `-march` flag to compile. We could probably use some IFUNC tricks to detect support at runtime with a fallback, which we would have to keep regardless for non-x86_64 architectures. I haven't really looked at AArch64 equivalents yet. In my opinion the complexity overhead of it is not really worth it and the standard implementation is already significantly faster than today (even with BMI2 disabled, bitset sampling becomes negligible in malariasimulation profiles).

For completeness, this is the same benchmark with more bitset sizes:
all-sizes

inst/include/IterableBitset.h
```diff
@@ -131,23 +133,37 @@ inline size_t popcount(uint64_t x) {
     #endif
 }
 
-//' @title find the next set bit
-//' @description given the current element p,
-//' return the integer represented by the next set bit in the bitmap
+//' Find the nth set bit in a 64bit integer.
+//'
+//' Returns the index of the bit, or 64 if there are not enough bits set.
+inline size_t find_bit(uint64_t x, size_t n) {
```
giovannic, line 139: This is the only concrete word type in the file, any reason?

why not `std::numeric_limits<A>::digits`?

giovannic (approved):
This is amazing stuff. Thank you!

inst/include/IterableBitset.h
plietar: I was mimicking the popcount and ctz functions above that hardcode to `uint64_t` because of the intrinsics they use. This was particularly relevant when this function was implemented using intrinsics itself.

I suppose I could make this function generic now that I'm not using intrinsics directly, but given that it calls ctz indirectly it wouldn't make much of a difference.

giovannic: Ah, now I see popcount and ctz. It was cut off from the github diff

PR #192, opened 2024-03-11
Add an all.equal implementation for Bitset.

plietar:
Calling `all.equal` (or equivalently `expect_equal`) on bitsets has never worked correctly. For some reason, externalptr always compare as equal, thus, under the R6 implementation, any bitset with the same max size was treated as equal.

With the recent switch to a named list and bitsets would sometimes be considered different, even where their contents were the same. More precisely, the arguments of `Bitset$new` now happen to be captured by all of the methods' environments and were being included in the comparison. A bitset created with `new(size=N)` would always be different compared to a bitset created with `new(from=ptr)`.

malariasimulation has some tests that use `mockery::expect_args` to compare bitset arguments, and these tests are now broken by the new individual version.


R/bitset.R
```diff
@@ -227,6 +227,17 @@ Bitset <- list(
   }
 )
 
+#' @export
+#' @method all.equal Bitset
+all.equal.Bitset <- function(target, current, ...) {
```
giovannic, line 232: looks like this method dispatches based on the `target` argument. Can we check the type of `current`?

giovannic (changes requested):
Sorry, I meant to request changes in my previous review

R/bitset.R
plietar: Done.

PR #194, opened 2024-03-22
Allow events and variables to be added when restoring the simulation.

plietar:
The simulation restoring feature required the number of events to be identical to the original simulation's. This has turned out to be too restrictive for our use cases in malariasimulation, where we want to resume the simulation with new interventions enabled, which come with new events.

I had originally hoped that this would be enough, and that we could run the first phase of the simulation with all the events present, even though they are unused. However there are interventions whose number of events depends on the parametrization of it, hence we cannot create a saved simulation state that fits all possible use cases when resuming.

The solution implemented here is to allow more events and variables to be introduced when resuming the simulation, on the condition that the objects are named when passed to `simulation_loop`. Without this requirement, given more objects than were saved, it would be ambigous which ones need to be restored and which ones are new.

Additionally, the list of events and variables can now be structured using nested lists of objects. This has no impact on the way the simulation is executed, but it allows for more complicated simulations to be restored.

For example, this is a simplified example of what the list of events in malariasimulation might look like:
```r
list(
 mda_administer = Event$new(),
 mass_pev_doses = list(
  TargetedEvent$new(n),
  TargetedEvent$new(n),
  TargetedEvent$new(n)
 )
)
```

In this example, the top-level list is names, allowing the `mass_pev_doses` events to be absent during the warmup simulation and added only later when restoring. However, because the nested list of targeted events is not named, more events cannot be added to it.

The names of the methods to save and restore state, together with the helper functions `save_object_state` and `restore_object_state` are tweaked and made public to allow this pattern to be reused in applications that have their own state to save, as demonstrated by the [stateful.R] file in malariasimulation.

[stateful.R]: https://github.com/mrc-ide/malariasimulation/blob/b3376d33cda29cc5e4d7217fefad4b367f9f9167/R/stateful.R


R/simulation.R
```diff
@@ -126,6 +139,54 @@ restore_state <- function(state, variables, events, restore_random_state) {
...
+#' @description Restore the state of one or more simulation objects. The
+#' specified objects are paired up with the relevant part of the state object,
+#' and the \code{restore_state} method of each object is called.
+#'
+#' If the list of object is named, more objects may be specified than were
+#' originally present in the saved simulation, allowing a simulation to be
+#' extended with more features upon resuming. In this case, the
+#' \code{restore_state} method is called with a \code{NULL} argument.
+#'
+#' @param objects a simulation object (ie. a variable or event), or list
```
giovannic, line 156: Could we have a description of `state`? It took me a while to figure out what the source and destination was.

And `timesteps` for completeness. CMD check will probably complain about this

R/simulation.R
```diff
@@ -126,6 +139,54 @@ restore_state <- function(state, variables, events, restore_random_state) {
...
+#' thereof.
+#' @export
+restore_object_state <- function(timesteps, objects, state) {
+  if (is.list(objects)) {
+    if (is.null(state)) {
+      keys <- NULL
+      reset <- seq_along(objects)
+    } else if (is_uniquely_named(objects) && is_uniquely_named(state)) {
+      missing <- setdiff(names(state), names(objects))
```
giovannic, line 165: Couldn't we issue a warning? We _could_ run a simulation with the given state.

giovannic (changes requested):
It looks like there's some renaming and minor refactoring too. Which looks fine.

R/event.R
```diff
@@ -40,11 +40,14 @@ Event <- R6Class(
   'Event',
   inherit = EventBase,
   public = list(
+    .should_restore = FALSE,
```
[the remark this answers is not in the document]
plietar: Done.

R/simulation.R
plietar: Done.

It does print some warnings about it, but there's already a couple dozen of those so I missed the new ones.

R/simulation.R
plietar: Yeah not really sure what to do about it. I agree it would totally be possible to run the simulation like that.

I don't think a warning is that useful, we should either support or not at all. Right now I don't really envision a use case where you would really need to remove objects - worst case you can always add them but not use them, or for example for an Event re-add previously scheduled timesteps that are all in the past, meaning the event never fires.

giovannic: Do we have to implement something to support it? Or not raise the error?

The worst case seems like a common case for malariasimulation. I imagine people will want to distribute historic runs with all the interventions turned on for general use. Then researchers interested in a specific intervention may want to focus on that. 

You think it's easier to add those empty events (and maybe variables I'm not sure) in malariasimulation so that people don't get errors when users continue a run?

It seemed easier to me for individual to ignore that state when restoring (probably even without a warning!), but I may be missing something here.

plietar: I've changed and removed that check. It now works even when adding and removing objects.

To be honest my plan with malariasimulation is to do the opposite, have a minimal historic run (that may include the  interventions that affect the past, but no more), and individual researchers can add more interventions. I guess we'll see if removing those interventions ever turns out to be useful.

giovannic (approved):

giovannic:
Oh, can you change the destination to dev before merging? Didn't notice that before I approved. Thanks!

PR #198, opened 2024-07-11
Allow processes to be named.

plietar:
Because of how R assigns names to stack frames, all processes in a typical individual simulation would end up being called `p`. This makes it difficult to interpret profiling results.

R uses the name of the variable the called function is bound to. By dynamically creating a variable with a chosen name and using `eval` to execute that variable, we can get the stack frame to show up with any desired name.

This uses this trick to allow the list of processes to be given names, and these names are used in the calls.

Below is a comparison of the before and after of looking at a malariasimulation profile. Instead of everything being grouped under a single `p` node, there are now distinct nodes for each process.

image
image

### DIFF
--- a/R/simulation.R
+++ b/R/simulation.R
@@ -57,9 +57,13 @@ simulation_loop <- function(
   flat_events <- unlist(events)
   flat_variables <- unlist(variables)
 
+  processes <- lapply(seq_along(processes), function(i) {
+    prepare_process(processes[[i]], names(processes)[[i]])
+  })
+
   for (t in seq(start, timesteps)) {
-    for (process in processes) {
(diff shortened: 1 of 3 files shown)


plietar:
Based on some quick local benchmarking there doesn't seem to be any measurable difference in performance with this.

I've kicked off a touchstone run on https://github.com/plietar/malariasimulation/pull/13, waiting for results.

giovannic (approved):
Thanks for figuring that out!

giovannic:
Note: we would want to update dev too.

plietar:
Yeah that was definitely aimed at dev, I keep forgetting to set the right target branch

PR #199, opened 2024-07-15
Support vectors of logicals as an argument to filter_bitset.

plietar:
In addition to a vector of numerical indices and another bitset, the filter_bitset now accepts a vector of logicals (ie. booleans). Each set bit of the bitset is included in the return value only if the corresponding value in the logical vector is TRUE.

Given `v` a logical vector, `filter_bitset(b, v)` is equivalent to the already supported `filter_bitset(v, which(v))` but is more performant. The new semantics also align with subsetting of R vectors, which similarly supports either a vector of indices or vector of logicals.

### DIFF
--- a/R/bitset.R
+++ b/R/bitset.R
@@ -246,20 +246,25 @@ all.equal.Bitset <- function(target, current, ...) {
 #' @description This non-modifying function returns a new \code{\link{Bitset}}
 #' object of the same maximum size as the original but which only contains
 #' those values at the indices specified by the argument \code{other}.
-#' Indices in \code{other} may be specified either as a vector of integers or as
-#' another bitset. Please note that filtering by another bitset is not a
-#' "bitwise and" intersection, and will have the same behavior as providing
-#' an equivalent vector of integer indices.
+#'
+#' Indices in \code{other} may be specified either as a vector of logicals, a
+#' vector of integers or as another bitset. If a vector of logicals is
+#' specified, it must be of the same size as the bitset. Please note that
+#' filtering by another bitset is not a "bitwise and" intersection, and will
+#' have the same behavior as providing an equivalent vector of integer indices.
 #' @param bitset the \code{\link{Bitset}} to filter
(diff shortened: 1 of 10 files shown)


tests/testthat/test-bitset.R
```diff
@@ -290,6 +290,20 @@ test_that("bitset filtering works for vector input", {
   expect_equal(filter_bitset(b, f)$to_vector(), c(1, 6))
 })
 
+test_that("bitset filtering works for logical input", {
+  b <- Bitset$new(10)$insert(c(1, 5, 6))
+  f <- c(TRUE, FALSE, TRUE)
+  expect_equal(filter_bitset(b, f)$to_vector(), c(1, 6))
+})
+
+test_that("bitset filtering works for logical input", {
```
giovannic, line 299:
```suggestion
test_that("bitset filtering errors when logical vector is too short", {
```

R/bitset.R
```diff
@@ -246,20 +246,25 @@ all.equal.Bitset <- function(target, current, ...) {
...
     if (other$size() > 0) {
       return(Bitset$new(from = filter_bitset_bitset(bitset$.bitset, other$.bitset)))
     } else {
       return(Bitset$new(size = bitset$max_size))
     }
+  } else if (inherits(other, "logical")) {
+      return(Bitset$new(from = filter_bitset_logical(bitset$.bitset, other)))
   } else {
     if (length(other) > 0) {
       return(Bitset$new(from = filter_bitset_vector(bitset$.bitset, as.integer(other))))
```
giovannic, line 270: I suggest that we rename filter_bitset_vector -> filter_bitset_integer.

```suggestion
  } else {
    # filter integer
    if (length(other) > 0) {
      return(Bitset$new(from = filter_bitset_integer(bitset$.bitset, as.integer(other))))
```

giovannic (changes requested):

R/bitset.R
plietar: Good point. Done.

giovannic (approved):

PR #200, opened 2024-07-15
Add a method to overwrite a bitset.

plietar:
This copies the value from one bitset to another one.

### DIFF
--- a/R/RcppExports.R
+++ b/R/RcppExports.R
@@ -41,8 +41,8 @@ bitset_or <- function(a, b) {
     invisible(.Call(`_individual_bitset_or`, a, b))
 }
 
-bitset_assign <- function(a, b) {
-    invisible(.Call(`_individual_bitset_assign`, a, b))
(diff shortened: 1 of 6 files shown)


plietar:
Somehow it hadn't even crossed my mind to just use `copy` for this. This does have a slight difference in behaviour in the case where the bitset is aliased, not that it matters much for the usecase I want it:

This will modify both x and y (since they are the same bitset):
```r
x <- Bitset$new(10)
y <- x
x$assign(z)
```

Where as this will only modify x, by creating a new bitset:
```r

x <- Bitset$new(10)
y <- x
x <- z$copy()
```

I still think this is useful to have. I'm not sure I love the name though. An alternative could be `copy_from`, inspired by Rust's [`clone_from`](https://doc.rust-lang.org/std/clone/trait.Clone.html).

I'll add the extra note to the docs you suggested.

plietar:
I've changed my mind and renamed to `copy_from`. Let me know what you think.

PR #201, opened 2024-07-16
Make num_bits a constant expression.

plietar:
The variable stores the number of bits per word and is used throughout the implementation to split a position into a word index and bit position in the word.

The value can be determined statically from the size of the template parameter and does not need to be stored as a field. Keeping it as a `constexpr` value allows the compiler to optimize operations that use it. In particular, since the value is always a power of two, any multiplication, division, and modulo operations by this value, which are quite common, can be turned into bit shifts or masks by a constant.

The performance impact of this change is quite small, but measurable, at around 2% for a 1M malariasimulation run, with no downside.


giovannic (approved):
<3

PR #206, opened 2025-03-14
Add bitset_count_and

cwhittaker1000:
A little while ago, @giovannic kindly helped me out with an issue I was facing and developed `bitset_count_and(a, b)` for a case in a model we've been developing. That function takes two bitsets and returns the size of the `$and`. This was motivated by a desire to avoid what we'd previously been doing, which looked like this: 
```
x$copy()$and(y)$size()
```
The copy was necessary as we needed to not modify `x`. But it became very computationally expensive as we were having to do this lots of times (lots of copying). This then became:
```
bitset_count_and(x, y)
```
which saved us a lot of computational overhead.

I'm conscious `individual` has moved on several versions since this was done. Would it be possible to think about merging this into main? So we can benefit from the addition of `bitset_count_and()` *and* the other updates to `individual`? Thanks in advance!


cwhittaker1000:
Tagging @giovannic and @plietar for visibility! :)

plietar:
This seems reasonable. Based on a quick search there's at least a handful of occurrences of this in malariasimulation. I'll try to spin up a PR there to see if there's any measurable impact.

Not sure I love the name though. The number of set bits is typically called `size`, not `count`, so I think we should stick to that. `bitset_intersection_size` seems more correct to me (but is a little more verbose).

What do you think?

inst/include/IterableBitset.h
```diff
@@ -79,6 +79,7 @@ class IterableBitset {
     IterableBitset& operator&=(const IterableBitset&);
     IterableBitset& operator|=(const IterableBitset&);
     IterableBitset& operator^=(const IterableBitset&);
+    size_t count_and(const IterableBitset&) const;
```
plietar, line 82: Doesn't make a big difference, but this feels like it should be a static method in the class:

```c++
static size_t count_and(const IterableBitset&, const IterableBitset&);
```

cwhittaker1000:
Wanted to circle back on this @plietar and @giovannic - we’re beginning to prep the work that uses this branch for publication so would love to merge this if you think that’d be possible!

Naming it bitset_intersection_size makes sense to me Paul! Apologies for not replying before (wasn’t sure whether it was my place to weigh in on names as I’ve not been involved in package dev) - though will defer to you and Giovanni re naming - whatever you think is best is good by me!

Thanks in advance folks!

cwhittaker1000:
Gentle nudge on this - @giovannic, do you think getting this merged might be possible?

giovannic (approved):
Good morning, I'll give this an approval and push this through since the tests appear to pass and the comments aren't severe.

I fixed up the documentation and and brought this up to date with dev to get the CI to pass.

Thanks and sorry for the delay on this!

cwhittaker1000:
Thanks so much @giovannic, hugely appreciated! :)

PR #211, opened 2026-06-08
R 4.6 compatibility updates:

giovannic:
* Make IterableBitset iterator compatible with C++20
 * Fix cheeky bug in `fast_bernoulli` where the probability member is unintentionally shadowed
 * Fix compiler warnings for fast bernoulli and (un)signed size comparisons in bitset api

### DIFF
--- a/inst/include/IterableBitset.h
+++ b/inst/include/IterableBitset.h
@@ -34,23 +34,21 @@ class IterableBitset {
     void unset(size_t);
     std::vector<A> bitmap;
 public:
-    using allocator_type = std::allocator<size_t>;
-    using value_type = allocator_type::value_type;
-    using reference = allocator_type::reference;
-    using const_reference = allocator_type::const_reference;
-    using difference_type = allocator_type::difference_type;
-    using size_type = allocator_type::size_type;
-    using input_iterator_type = std::iterator<std::input_iterator_tag, size_t>;
+    using value_type = size_t;
+    using reference = size_t&;
+    using const_reference = const size_t&;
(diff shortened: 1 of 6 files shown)


inst/include/IterableBitset.h
```diff
@@ -34,23 +34,21 @@ class IterableBitset {
...
-    using reference = allocator_type::reference;
-    using const_reference = allocator_type::const_reference;
-    using difference_type = allocator_type::difference_type;
-    using size_type = allocator_type::size_type;
-    using input_iterator_type = std::iterator<std::input_iterator_tag, size_t>;
+    using value_type = size_t;
+    using reference = size_t&;
+    using const_reference = const size_t&;
+    using difference_type = std::ptrdiff_t;
+    using size_type = std::size_t;
```
M-Kusumgar, line 41: Just for my understanding, why were you allocating vs not now?

M-Kusumgar (approved):
All seems reasonable, nice bug fix as well, I'm a bit unsure about the allocator removal or even why you were allocating in the first place, I think I'm missing some context but otherwise all looks good!

inst/include/IterableBitset.h
giovannic: This was written when C++14 was the default and std::allocator members were a thing. In c++17 they were [deprecated](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2016/p0174r2.html#2.2) because they are redundant. And in C++20 they were [removed](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2018/p0619r3.html#3.9)

